PR #22, opened 2015-01-03
Adds auto-scaling to LinearSVC and adds psudeo-Feature modifiers.log

halfak:


revscores/scorers/linear_svc.py
```diff
@@ -74,7 +78,20 @@ def test(self, values_scores):
...
+        columns = zip(*values)
+        
+        stats = tuple((mean(c), stdev(c)) for c in columns)
         
+        return stats
+    
+    def _scale_and_center(self, values, stats):
+        
+        for feature_values in values:
+            yield tuple((val-mean)/max(sd, 0.01)
```
he7d3r: Is this division by 0.01 a good default when all the values are identicall (sd = 0)?

halfak: Well, in that case, all val-mean would be zero and zero divided by 0.01 is zero.  Generally, features that are all identical aren't good predictors.

he7d3r: Oh, off course! Looks good then.

halfak: Anything else you want me to take a look at?

halfak:
Merge

halfak:
FYI: Accidentally clicked the close button.  I reopened the pull request.

revscores/scorers/linear_svc.py
he7d3r: It is ok for now.

PR #24, opened 2015-01-10
Refactors language, adds tests and updates demos/examples

halfak:


revscores/languages/english.py
```diff
@@ -27,6 +27,8 @@
     "fart",
     "fat",
     "fuck",
+    "fucker",
+    "fucking",
```
he7d3r: This new word is unnecessary:
```python
>>> STEMMER.stem("fucking")==STEMMER.stem("fuck")
True
```

revscores/languages/language.py
```diff
@@ -0,0 +1,19 @@
+
```
he7d3r: Do we have conventions for whitespaces, linebreaks, etc?
This line 1 seems unnecessary.

revscores/languages/language.py
```diff
@@ -0,0 +1,19 @@
...
+class Language:
+    """
+    Constructs a Language instance that wraps two functions, "is_badword" and "is_misspelled" -- which will
+    check if a word is "bad" or does not appear in the dictionary.
+    """
+    def __init__(self, is_badword, is_misspelled):
+        self.is_badword = is_badword
+        self.is_misspelled = is_misspelled
+    
+    def badwords(self, words):
```
he7d3r: Do we need to document that this takes a list of words and returns the sublist of those which are bad? (same question below, about misspellings)

revscores/languages/portuguese.py
```diff
@@ -1064,21 +1064,20 @@
...
-                if len(wordnet.synsets(word, lang="por")) == 0:
-                    yield word
+        return len(wordnet.synsets(word, lang="por")) == 0
+
+portuguese = Language(
+    is_badword,
+    is_misspelled
+)
+portuguese.STEMMER = STEMMER
+portuguese.BADWORDS = BADWORDS
```
he7d3r: Are there conventions about the use of upper case letters in variable names?

revscores/languages/portuguese.py
```diff
@@ -1064,21 +1064,20 @@
     "zueira"
 ])
 
-class Portuguese(Language):
+def is_badword(word):
+    return STEMMER.stem(word).lower() in BADWORDS
```
he7d3r: Considering that the definition of `is_badword` is the same for Portuguese and English, isn't possible to make this to be a "default implementation" so that we don't need to have two (or more) copies of it?

revscores/languages/portuguese.py
```diff
@@ -1064,21 +1064,20 @@
...
+                
+def is_misspelled(word):
+    if word in ("a", "A"): return False
```
he7d3r: This is hacky. Is "a" the only (single letter) word which causes problem?
("e" and "o" are also words in Portuguese)

revscores/languages/tests/test_english.py
```diff
@@ -0,0 +1,19 @@
...
+    assert not english.is_badword("hat")
+    
+    assert english.is_misspelled("wjwkjb")
+    assert not english.is_misspelled("waffles")
+    
+    eq_(list(english.badwords(["good", "slut", "bad", "ugly", "fucker"])),
+        ["slut", "fucker"])
+    
+    eq_(list(english.misspellings(["waffles", "oof", "dog", "oof"])),
+        ["oof", "oof"])
```
he7d3r: Is it useful to keep the list with duplicate items?

revscores/languages/portuguese.py
halfak: Indeed.  Upper case used with constants "often set at the module level".   https://www.python.org/dev/peps/pep-0008/#constants

revscores/languages/portuguese.py
halfak: Gotcha.  I'll add "o" for the time being.  We should probably open new tickets for new bugs :)

revscores/languages/portuguese.py
halfak: It's weird to generalize the is_badword() strategy without generalizing is_misspelled().  I considered it, but it ends up requiring a more complex class structure.  I don't think it's worth it.

PR #26, opened 2015-01-17
Provided code from revscores hacking with Halfak

ToAruShiroiNeko:


revscores/scorers/tests/test_linear_svc.py
```diff
@@ -33,8 +31,17 @@ def test_model():
...
+
+    def test_serialization():
+        model = LinearSVC.MODEL([proportion_of_badwords_added,
+                                 proportion_of_symbolic_added])
+        f = BytesIO()
+        model.dump(f)
+        f.seek(0) # Rewind the file
+        reconstructed_model = MLScorerModel.load(f)
+        eq_(reconstructed_model.features, model.features)
+        eq_(type(reconstructed_model), type(model))
```
halfak: This shouldn't be tabbed in like this.  It needs to be tabbed back to zero in order to run the actual test.

revscores/scorers/tests/test_linear_svc.py
```diff
@@ -1,14 +1,12 @@
 from io import BytesIO
-
+from revscoring.scorers import MLScorerModel
```
halfak: should be "from ..scorer import MLScorerModel"

PR #31, opened 2015-01-20
Update Ipython notebook

he7d3r:
Follows up 28bb1dc7dae0a48b3316c670282c5a3c83608cda.

### DIFF
--- a/ipython/Demo_LinearSVC.ipynb
+++ b/ipython/Demo_LinearSVC.ipynb
@@ -113,7 +113,7 @@
        "output_type": "stream",
        "stream": "stdout",
        "text": [
-        "{'seconds_elapsed': 1.739048957824707}\n"
+        "{'seconds_elapsed': 1.649972915649414}\n"
        ]
       }
      ],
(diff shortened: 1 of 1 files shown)


he7d3r:
Should I also update the
`"signature": "sha256:4293dfc1fa813e265ed74d298f9adcf8a043a94e827a714142d1a7acf4b12199"`
? To what?

he7d3r:
@halfak : what about the signature (question above)?

halfak:
Sorry, I'm not sure what signature you are referring to.  This might be something internal to ipython notebook's format.  I think we ought to focus on the visual presentation of the ipython notebook and not concern with it's underlying json representation.

he7d3r:
@halfak: But how did create the notebook in the first place if not by editing the JSON (and adding the existing signature manually)?

halfak:
Oh!  "ipython" comes with a "notebook" for editing.  This "notebook" starts up a minimal webserver and gives you an editing interface in your browser. 

1. Install ipython with "pip install ipython"
2. Start the notebook server with "ipython notebook" 
2.5 This should automatically pull up a browser window for you, but if not, go to "http://localhost:8888"
3. Create and edit "notebooks" that exist in the directory from which you ran "ipython notebook"

he7d3r:
Good to know. Thanks!

PR #35, opened 2015-01-29
Adds statistics to requirements

halfak:
Added 'statistics' to requirements for compatibility with python 3.x where x < 4

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ numpy
 pytz
 scikit-learn
 scipy
+statistics


halfak:
FYI, I checked and this seems to cause no harm for python 3.4 installs.

he7d3r:
`$ ack-grep revscores`
still returns a few results (mostly in examples).

setup.py
```diff
@@ -8,11 +8,11 @@ def read(fname):
...
 
 def requirements(fname):
-    return [line.strip() 
+    return [line.strip()
             for line in open(os.path.join(os.path.dirname(__file__), fname))]
 
 setup(
-    name="revscores",
+    name="revscoring",
     version=read('VERSION').strip(),
```
he7d3r: Consider increasing the version number from 0.0.2.

halfak: The only time we should be increasing version numbers is when we publish a release ( i.e. upload a snapshot to PyPI)

PR #39, opened 2015-01-31
Release work

halfak:
Preparing for a PyPI release.  Cleans up readme.rst and adds an example use of Scorer.


README.rst
```diff
@@ -76,6 +72,37 @@ Feature extraction:
...
+    >>> Downloader> d
+    >>> Identifier> wordnet
+    >>> Downloader> d
+    >>> Identifier> omw
+    >>> Downloader> q
+    >>> exit()
+
+
+You might need to install some other dependencies depending on your operating
+system.  These are for ``scipy`` and ``numpy``.
```
he7d3r: I'm not sure `python-sklearn` is for ``scipy`` and ``numpy`` too.

example_scoring.py
```diff
@@ -0,0 +1,17 @@
+from mw.api import Session
+
+from revscoring.extractors import APIExtractor
+from revscoring.languages import english
+from revscoring.scorers import MLScorerModel
+
+api_session = Session("https://en.wikipedia.org/w/api.php")
+extractor = APIExtractor(api_session, english)
+
+filename = "models/reverts.halfak_mix.trained.model"
```
he7d3r: Maybe add a comment mentioning that this file would have been generated by some other command (linking to an example?)

README.rst
```diff
@@ -4,41 +4,37 @@ A generic, machine learning-based revision scoring system designed to be used
...
+        >>>
+        >>> from revscoring.extractors import APIExtractor
+        >>> from revscoring.languages import english
+        >>> from revscoring.scorers import MLScorerModel
+        >>>
+        >>> api_session = Session("https://en.wikipedia.org/w/api.php")
+        Sending requests with default User-Agent.  Set 'user_agent' on api.Session to quiet this message.
+        >>> extractor = APIExtractor(api_session, english)
+        >>>
+        >>> filename = "models/reverts.halfak_mix.trained.model"
```
he7d3r: Maybe add a comment mentioning that this file would have been generated by some other command (linking to an example?)

README.rst
halfak: Looks like scipy and numpy are the only ones that suffer from those dependencies.  See https://meta.wikimedia.org/wiki/Research_talk:Revision_scoring_as_a_service#Installing_and_checking_requirements.txt

example_scoring.py
halfak: I think that is fair and we should actually have this example added at some point.  Right now, I don't think we have a good place to link to.  And I expect that most of our users won't be building their own models, so this is a more important example to have now.

README.rst
halfak: See note on example_scoring.py

README.rst
he7d3r: I mean, the comment says "g++ git python-sklearn python3.4-dev libatlas-base-dev gfortran" are for `scipy` and `numpy`, but you were able to install them before `scikit-learn`, so `python-sklearn` is not needed for them (but for sklearn itself).

halfak: Oh!  Yeah, it looks like "python-sklearn" came from one of *your* pull requests.  :P  I'm happy to remove it.

he7d3r: Ok. I'll let it like this for now and check again later...

PR #44, opened 2015-02-01
Generalizes MLScorerModel, adds languages to model.  Updates tests and examples. 

halfak:


halfak:
This is related to https://github.com/halfak/Revision-Scoring/issues/38

revscoring/scorers/svc.py
```diff
@@ -41,9 +42,11 @@ def train(self, values_scores, balanced_weight=True):
             counts = {}
             for score in scores:
                 counts[score] = counts.get(score, 0) + 1
-
-            weights = [1/(counts[s]/len(scores)) for s in scores]
-            self.svc.fit(scaled_values, scores, weights)
+            
+            self.weights = {s:1-(counts[s]/len(scores)) for s in counts}
```
he7d3r: I'm not sure I understand these changes.

revscoring/scorers/svc.py
```diff
@@ -51,52 +54,73 @@ def train(self, values_scores, balanced_weight=True):
...
         """
-        values, scores = zip(*values_scores)
+        values, labels = zip(*values_labels)
+        values, labels = list(values), list(labels)
+        
+        scores = list(self.score(values))
+        
+        if comparison_class == "auto":
+            comparison_class = self.svc.classes_[1]
+        elif comparison_class not in self.svc.classes_[1]:
```
he7d3r: You sure about these two lines above? It seems `self.svc.classes_[1]` is either a class (so you can assign it to `comparison_class`) or composed of many classes (so that you can check if `comparison_class` is `in` it). Which one is correct? Or did I miss something?

revscoring/scorers/svc.py
```diff
@@ -51,52 +54,73 @@ def train(self, values_scores, balanced_weight=True):
...
         
-        scaled_values = list(self._scale_and_center(values, self.feature_stats))
         
-        true_probas = [p[1] for p in self.svc.predict_proba(scaled_values)]
-        fpr, tpr, thresholds = roc_curve(scores, true_probas)
+        probabilities = [s['probability'][comparison_class]
+                         for s in scores]
+        predicteds = [s['prediction'] for s in scores]
+        
+        predicted_true = [p == comparison_class for p in predicteds]
```
he7d3r: Is `predicted_true` supposed to be unused somewhere?

revscoring/scorers/tests/test_svc.py
```diff
@@ -1,34 +1,56 @@
...
+    observations = list(chain(
+        zip(((some, other) for some, other in
+             zip((deterministic.normalvariate(1, .3) for i in range(500)),
+                 (deterministic.normalvariate(2, .5) for i in range(500)))),
+            (True for i in range(100))),
+        zip(((some, other) for some, other in
+             zip((deterministic.normalvariate(-1, .5) for i in range(35)),
+                 (deterministic.normalvariate(-2, .3) for i in range(35)))),
+            (False for i in range(100)))
+    ))
```
he7d3r: This syntax confuses me a little, so ignore me in case the question is stupid: aren't these `range(100)` supposed to be the same size as those `range(500)` and `range(35)`?

revscoring/scorers/svc.py
halfak: It makes self.weights accessible later.  This may be useful for re-scaling probability estimates.  Regardless, the weights we use when training should be observable.

revscoring/scorers/svc.py
halfak: Bah!  Good catch.  It is should be "not in self.svc.classes_"

revscoring/scorers/svc.py
halfak: Nope.  Looks like we don't need it.

revscoring/scorers/tests/test_svc.py
halfak: Yes they are!  Good catch.

PR #48, opened 2015-02-12
Fixes JSONable bug in SVC scores output and adds a test to show that it works

halfak:
Fixes https://github.com/halfak/Revision-Scoring/issues/47


revscoring/scorers/tests/test_svc.py
```diff
@@ -45,7 +46,11 @@ def train_score(model):
     
     model.train(train_set)
     score_doc = next(model.score([(-1,-2)]))
+    
     eq_(score_doc['prediction'], False)
+    print(score_doc)
+    print(type(list(score_doc['probability'].keys())[0]))
```
he7d3r: Are these two `print()` calls supposed to be here?

revscoring/scorers/util.py
```diff
@@ -0,0 +1,20 @@
...
+def normalize(v):
+    if isinstance(v, numpy.bool_):
+        return bool(v)
+    else:
+        return v
+
+def normalize_json(doc):
+    
+    if isinstance(doc, dict):
+        return {normalize_json(k):normalize_json(v) for k,v  in doc.items()}
```
he7d3r: Does this work recursivelly? E.g. in objects like `{a:1, b:{c:0, d:2} }`, will 0 and 2 be treated as items?

halfak: Yes.  It seems to.  I've added a test to make sure.

PR #55, opened 2015-02-23
Fixes bug in dependency solver cache.

halfak:
The old dependency solver would repeat computations when solving multiple dependencies.  

This commit adds solve_many() to preserve the cache and a test to make sure that it works as expected.

### DIFF
--- a/revscoring/dependent.py
+++ b/revscoring/dependent.py
@@ -12,9 +12,11 @@ class Dependent:
         self.name = name
         self.process = process
         self.dependencies = dependencies if dependencies is not None else []
+        self.calls = 0
     
     def __call__(self, *args, **kwargs):
         logger.debug("Executing {0}.".format(self))
+        self.calls += 1
         return self.process(*args, **kwargs)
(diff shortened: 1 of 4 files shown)


revscoring/dependent.py
```diff
@@ -12,9 +12,11 @@ def __init__(self, name, process, dependencies=None):
         self.name = name
         self.process = process
         self.dependencies = dependencies if dependencies is not None else []
+        self.calls = 0
     
     def __call__(self, *args, **kwargs):
         logger.debug("Executing {0}.".format(self))
+        self.calls += 1
```
he7d3r: Is there any chance of this hitting the max allowed integer (is there such a thing in Python?) when processing a huge dataset? Would that cause an error?

halfak: Indeed, that is possible, but Python gracefully transitions to using longs when necessary.

See http://stackoverflow.com/questions/13795758/what-is-sys-maxint-in-python-3

PR #56, opened 2015-02-23
Refactors language utilities to work as dependencies as well.

halfak:
This is a big change.  The primary difference is that Language is refactored to provide a set of dependencies to inject when solving for features.  This will allow languages to provide a specific subset of utilities (e.g. is_badword) and work with features that require only those utilities that are implemented.  

Otherwise, I took the opportunity to simplify some of the compound features and clean up the test cases.


revscoring/datasources/previous_user_rev_doc.py
```diff
@@ -7,10 +7,10 @@ def process(session, revision_metadata):
...
     if revision_metadata.user_text is not None:
         docs = session.user_contribs.query(user={"EpochFail"},
-                                          properties={'ids'},
-                                          limit=1,
-                                          direction="older",
-                                          start="20140901010101")
+                                           properties={'ids','timestamp'},
+                                           limit=1,
+                                           direction="older",
+                                           start=revision_metadata.timestamp-1)
```
halfak: Assumes that a user can't save two revisions in the same second.  While this is not idea, it seems to work out in practice.  http://blog.wikimedia.org/2014/12/29/how-we-made-editing-wikipedia-twice-as-fast/

PR #57, opened 2015-02-24
Scoring refactor

halfak:
model.score() now only works with one set of feature_values at a time.  This works better with the realities of gathering features and the errors that might occur.  

This also makes it so that a Scorer can accept multiple ScorerModels.  The Scorer will then extract features efficiently for all of the models that it was configured with.  

Test cases, demos and examples have been updated.


revscoring/scorers/scorer.py
```diff
@@ -1,141 +1,320 @@
...
-                for rev_id in rev_ids)
+        section = config['scorers'][key]
         
-    
+        model_map = {}
+        for name, key in section['models']:
+            model = ScorerModel.from_config(config, key)
+            model_map[name] = model
+        
+        extractor = Extractor.from_config(config, section['extractor'])
```
he7d3r: What is this "Extractor.from_config"?

he7d3r: Reported as https://github.com/halfak/Revision-Scoring/issues/64

revscoring/scorers/scorer.py
```diff
@@ -1,141 +1,320 @@
...
+    def _scale_and_center(self, values, stats):
         
+        for feature_values in values:
+            yield (tuple((val-mean)/max(sd, 0.01)
+                   for (mean, sd), val in zip(stats, feature_values)))
     
-    def test(self, values_scores):
+    @classmethod
+    def from_config(cls, config, key):
+        section = config['m    odels'][key]
```
he7d3r: typo: `'m odels'`

revscoring/scorers/scorer.py
```diff
@@ -1,141 +1,320 @@
...
+        :Returns:
+            A dictionary with the fields:
+            
+            * seconds_elapsed -- Time in seconds spent fitting the model
+        """
+        start = time.time()
+        
+        values, labels = zip(*values_labels)
+        
+        # Fit SVC model
```
he7d3r: This comment looks outdated: `classifier_model` might not be an SVC model, right?

PR #59, opened 2015-02-28
Adding Azerbaijani language

ToAruShiroiNeko:


revscoring/languages/azerbaijani.py
```diff
@@ -0,0 +1,27 @@
...
+
+# Notice:
+# It seems like Azerbaijani stemmer isn't in the nltk snowball library for some reason:
+# http://www.nltk.org/_modules/nltk/stem/snowball.html
+# STEMMER = SnowballStemmer("azerbaijani")
+
+#STOPWORDS = set([])
+BADWORDS = set([])
+
+def is_badword_process():
```
halfak: We shouldn't provide this function if it won't work.  It's much better to leave it a missing dependency.

revscoring/languages/azerbaijani.py
```diff
@@ -0,0 +1,27 @@
...
+    def is_badword(word):
+        return word.lower() in BADWORDS
+    return is_badword
+is_badword = LanguageUtility("is_badword", is_badword_process, depends_on=[])
+
+def is_stopword_process():
+    def is_stopword(word):
+        return word.lower() in STOPWORDS
```
halfak: This line should fail since STOPWORDS doesn't exist.

halfak:
Bugfix commit works for me.

PR #60, opened 2015-03-04
Feature refactor and test expansion

halfak:
Also, adds features that are useful for Wiki-Class.  

The demos have been updates (we now have ~100 features!).  The tests are also complete.  All non-Modifier Features have their own tests. :)


he7d3r:
I'm getting this when running the nosetests for this pull request:
`ERROR: Failure: ImportError (No module named 'mwparserfromhell')`
It seems we have a new dependency, which should be added to the list.

he7d3r:
Is there a way (and is it necessary?) to specify that our package requires a specific version of another package?
(I have in mind the case of 'deltas')

README.rst
```diff
@@ -40,21 +40,21 @@ Feature extraction:
...
         >>> api_extractor = APIExtractor(Session("https://en.wikipedia.org/w/api.php"))
         Sending requests with default User-Agent.  Set 'user_agent' on api.Session to quiet this message.
         >>>
-        >>> features = [bytes_changed, chars_added, day_of_week_in_utc,
-        ...               hour_of_day_in_utc, is_custom_comment, user_age_in_seconds,
-        ...               user_is_anon, user_is_bot]
+        >>> features = [revision.day_of_week,
+        ...             revision.hour_of_day,
+        ...             revision.has_custom_comment,
+        ...             parent_revision.bytes_changed,
```
he7d3r: This looks weird... Isn't the number of bytes changed something related to the diff instead of only to the parent_revision?

README.rst
```diff
@@ -63,14 +63,15 @@ Feature extraction:
...
-        <day_of_week_in_utc>: 6
-        <hour_of_day_in_utc>: 19
-        <is_custom_comment>: True
-        <user_age_in_seconds>: 71821407
-        <user_is_anon>: False
-        <user_is_bot>: False
+        <revision.day_of_week>: 6
+        <revision.hour_of_day>: 19
+        <revision.has_custom_comment>: True
+        <(revision.bytes - parent_revision.bytes_changed)>: 3
```
he7d3r: Wouldn't it be clearer if this line were
`<(revision.bytes - parent_revision.bytes)>: 3`
(this difference is what I would call the number of "bytes changed" in a diff, and might be the reason why the subtraction in line 69 looks weird to me)

README.rst
halfak: Good Q.  It's actually a byte length subtraction which does not require a diff.  BUT you've pointed out a problem with it below that I'll address.

he7d3r: I mean, a subtraction requires two values, one from the parent revision and one from the current revision, so it is more a property of the diff (the pair of revisions) than of the parent revision alone.

halfak: I struggle with this feature.  If you feel strongly, we can move it into the diff module.  The primary reason why I want to keep it out of the diff module is that it bears no relationship with the diff -- just a comparison of metadata items.  However, I agree that it *sounds* like a diff feature.

README.rst
halfak: Yes.  Fixed in last push.

README.rst
he7d3r: We can keep it like this.

halfak: Bah.  Now I have changed my mind.  I'm convinced that it will do no harm to move this to diff.  I've just updated the code to do so.

he7d3r: LOL

revscoring/datasources/site.py
```diff
@@ -0,0 +1,32 @@
...
+                                                  'namespacealiases'})
+        return doc
+
+doc = Datasource("site.doc", process_doc, depends_on=['session'])
+
+
+def process_namespace_map(site_info_doc):
+    
+    aliases = site_info_doc.get('namespacealiases', [])
+    alias_map = {}
```
he7d3r: This fixes https://github.com/halfak/Revision-Scoring/commit/2d6b636b16b7760a5f904c3b83b835251d638b8b#diff-288091a7883dc67d96ad359400b9ac0dR10

Thanks! :-)

halfak: Horray for tests pointing out old bugs! :S

revscoring/datasources/revision.py
```diff
@@ -0,0 +1,83 @@
...
+internal_links = Datasource("revision.internal_links", process_internal_links,
+                            depends_on=[parse_tree])
+
+def process_tags(revision_parse_tree):
+    return revision_parse_tree.filter_tags()
+
+tags = Datasource("revision.tags", process_tags, depends_on=[parse_tree])
+
+
+def profess_templates(revision_parse_tree):
```
he7d3r: profess?

revscoring/datasources/revision.py
```diff
@@ -0,0 +1,83 @@
...
+    return revision_parse_tree.filter_templates()
+
+templates = Datasource("revision.templates", profess_templates,
```
he7d3r: profess?

revscoring/datasources/tests/test_previous_user_revision.py
```diff
@@ -0,0 +1,33 @@
+from deltas import Delete, Equal, Insert
```
he7d3r: Unused?

revscoring/features/parent_revision.py
```diff
@@ -0,0 +1,118 @@
...
+################################## Bytes #######################################
+
+def process_bytes(parent_revision_metadata):
+    return parent_revision_metadata.bytes \
+           if parent_revision_metadata is not None else 0
+
+bytes = Feature("parent_revision.bytes", process_bytes,
+                returns=int, depends_on=[parent_revision.metadata])
+
+def process_was_same_user(parent_revision_metadata, revision_metadata):
```
he7d3r: This is another place where the property sounds more natural if associated to the pair of revisions instead of a single (the parent) revision. In the end it is using the metadata from both revisions...
(not a blocker though)

halfak: Any suggestions on how you'd rather it be organized?

halfak:
@he7d3r can this be merged?

revscoring/features/revision.py
```diff
@@ -0,0 +1,235 @@
...
+
+hour_of_day = Feature("revision.hour_of_day", process_hour_of_day,
+                      returns=int, depends_on=[revision.metadata])
+
+################################ Comment #######################################
+
+def process_has_custom_comment(revision_metadata):
+    
+    if revision_metadata.comment is not None:
+        trimmed_comment = SECTION_COMMENT_RE.sub("", revision_metadata.comment)
```
he7d3r: Does this handle multiple sections in a single edit summary? E.g. ""/*test*/ some comment /* another test*/""...

revscoring/features/revision.py
```diff
@@ -0,0 +1,235 @@
...
+                returns=int, depends_on=[revision.headings])
+
+def process_level_6_headings(headings):
+    return sum(h.level==5 for h in headings)
+
+level_6_headings = \
+        Feature("revision.level_6_headings", process_level_6_headings,
+                returns=int, depends_on=[revision.headings])
+
+def process_infonoise(is_stopword, stem_word, content_words):
```
he7d3r: Why are the non-stopwords called noise? Isn't it the opposite (the stopwords are the noise)?

revscoring/features/util.py
```diff
@@ -0,0 +1,11 @@
+import re
+
+CATEGORY_RE = re.compile(r"^category:", re.U|re.I)
+CITE_RE = re.compile(r"cite", re.I|re.U)
+IMAGE_RE = re.compile(r"^(file|image):", re.U|re.I)
+INFOBOX_RE = re.compile(r"infobox|sidebar", re.I|re.U)
```
he7d3r: The regexes CATEGORY_RE, CITE_RE, IMAGE_RE and INFOBOX_RE should be configurable per language (they are English only for now)

revscoring/features/parent_revision.py
he7d3r: "seconds_since" and "was_same_user" sound like properties of the diff, not of the parent_revision

revscoring/features/tests/test_revision.py
```diff
@@ -0,0 +1,292 @@
...
+    }
+    eq_(solve(infonoise, cache=cache), 2/12)
+
+def test_category_links():
+    FakeLink = namedtuple("FakeLink", "title")
+    cache = {
+        revision.internal_links: [FakeLink("Foobar"),
+                                  FakeLink("File:Hat.jpg"),
+                                  FakeLink("file:Bum.gif"),
+                                  FakeLink("Category:Hats")]
```
he7d3r: Once we have regexes per language, we would need a test for e.g. "Categoria:Matemática" in Portuguese...

revscoring/features/tests/test_revision.py
```diff
@@ -0,0 +1,292 @@
...
+    }
+    eq_(solve(category_links, cache=cache), 1)
+
+def test_image_links():
+    FakeLink = namedtuple("FakeLink", "title")
+    cache = {
+        revision.internal_links: [FakeLink("Foobar"),
+                                  FakeLink("File:Hat.jpg"),
+                                  FakeLink("file:Bum.gif"),
```
he7d3r: What about namespace aliases? Do they work too? E.g.: "File" vs "Image".

Issue #69, opened 2015-03-27
Bug from refractoring: max() arg is an empty sequence


ToAruShiroiNeko:
```python
....Traceback (most recent call last):
  File "/home/eva/Github/Objective-Revision-Evaluation-Service/ores/features_reverted.py", line 94, in run
    print('\t'.join(str(v) for v in (list(values) + [reverted])))
  File "/usr/local/lib/python3.4/dist-packages/revscoring-0.1.0-py3.4.egg/revscoring/dependent.py", line 34, in solve_many
    value, cache, history = _solve(dependent, cache)
  File "/usr/local/lib/python3.4/dist-packages/revscoring-0.1.0-py3.4.egg/revscoring/dependent.py", line 91, in _solve
    value, cache, history = _solve(dependency, cache, history)
  File "/usr/local/lib/python3.4/dist-packages/revscoring-0.1.0-py3.4.egg/revscoring/dependent.py", line 91, in _solve
    value, cache, history = _solve(dependency, cache, history)
  File "/usr/local/lib/python3.4/dist-packages/revscoring-0.1.0-py3.4.egg/revscoring/dependent.py", line 95, in _solve
    value = dependent(*values)
  File "/usr/local/lib/python3.4/dist-packages/revscoring-0.1.0-py3.4.egg/revscoring/features/feature.py", line 31, in __call__
    value = super().__call__(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/revscoring-0.1.0-py3.4.egg/revscoring/dependent.py", line 20, in __call__
    return self.process(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/revscoring-0.1.0-py3.4.egg/revscoring/features/diff.py", line 137, in process_longest_repeated_char_added
    for segment in diff_added_segments
ValueError: max() arg is an empty sequence
```

halfak:
Resolved by related pull request.

PR #73, opened 2015-04-03
Adds tests to detect #69 and updates code to fix the bug.

halfak:
What it says on the tin.

### DIFF
--- a/revscoring/features/diff.py
+++ b/revscoring/features/diff.py
@@ -7,8 +7,6 @@ from ..languages import is_badword, is_misspelled
 from .feature import Feature
 from .util import MARKUP_RE, NUMERIC_RE, SYMBOLIC_RE
 
-
-
 bytes_changed = revision.bytes - parent_revision.bytes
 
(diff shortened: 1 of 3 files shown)


revscoring/features/diff.py
```diff
@@ -133,10 +131,13 @@ def process_uppercase_chars_removed(diff_removed_segments):
...
-                    for segment in diff_added_segments
-                    for _, group in groupby(segment.lower()))
-
+    try:
+        return max(sum(1 for _ in group)
+                   for segment in diff_added_segments
+                   for _, group in groupby(segment.lower()))
+    except ValueError:
+        # Happens when there's no segments added
+        return 0
```
he7d3r: Isn't 1 a better value than 0 in this case? E.g. If I add "foo", the longest repeated char added was "o", which appeared 2 times. However, if I add "bar", each of the characters was added exactly 1 time.

halfak: Either value is sort of nonsensical.  I suppose that in the cases where there actually was added content, the min expected value is 1.  I don't have a strong opinion either way.  It seems like you'd prefer 1, so I'll make the change.

PR #75, opened 2015-04-03
Adds content_chars feature for use in wikiclass.

halfak:
Includes additional feature code and tests.


halfak:
Just realized that I forgot to update the demo code.  I'll do that quick and push another commit.

PR #84, opened 2015-04-12
PEP8

Ladsgroup:
Part III

### DIFF
--- a/revscoring/features/diff.py
+++ b/revscoring/features/diff.py
@@ -18,19 +18,22 @@ def process_segments_added(diff_added_segments):
 segments_added = Feature("diff.segments_added", process_segments_added,
                          returns=int, depends_on=[diff.added_segments])
 
+
 def process_segment_removed(diff_segments_removed):
     return len(diff_segments_removed)
 
 segments_removed = Feature("segments_removed", process_segment_removed,
                            returns=int, depends_on=[diff.removed_segments])
 
-############################## Characters ######################################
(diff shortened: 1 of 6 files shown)


revscoring/features/diff.py
```diff
@@ -39,96 +42,106 @@ def process_chars_removed(diff_removed_segments):
                         returns=int, depends_on=[diff.removed_segments])
 
 proportion_of_chars_removed = chars_removed / \
-              modifiers.max(parent_revision.chars, 1)
+    modifiers.max(parent_revision.chars, 1)
```
halfak: This is supposed to be tabbed in at least two levels.  
https://www.python.org/dev/peps/pep-0008/#indentation

revscoring/features/previous_user_revision.py
```diff
@@ -4,13 +4,13 @@
 
 def process_seconds_since(pur_metadata, revision_metadata):
     revision_timestamp = revision_metadata.timestamp \
-                         if revision_metadata is not None and \
-                            revision_metadata.timestamp is not None \
-                         else Timestamp(0)
+        if revision_metadata is not None and \
```
halfak: The old indents lined up this code in a nice way.

revscoring/features/diff.py
Ladsgroup: it's when the next line needs to be tabbed once (like for, if, function definition, etc)  so we can distinguish between logical lines.

PR #89, opened 2015-04-21
Fixes bugs in from_config() for Language and APIExtractor

halfak:


revscoring/scorers/scorer.py
```diff
@@ -68,8 +69,8 @@ def from_config(cls, config, name, section_key="scorers"):
                     extractor: ptwiki
 
             extractors:
-                enwiki: ...
-                ptwiki: ...
+                enwiki_api: ...
+                ptwiki_api: ...
 
             models:
```
he7d3r: `models` --> `scorer_models`?

PR #90, opened 2015-04-26
Add docs for features

Ladsgroup:
Part I

### DIFF
--- a/revscoring/features/diff.py
+++ b/revscoring/features/diff.py
@@ -31,17 +31,73 @@ def process_chars_added(diff_added_segments):
 
 chars_added = Feature("diff.chars_added", process_chars_added,
                       returns=int, depends_on=[diff.added_segments])
+"""
+Represents number of characters added in this edit.
+
+:Returns:
+    int
+
+:Example:
+    ..code_block :: python
+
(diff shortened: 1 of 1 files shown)


revscoring/features/diff.py
```diff
@@ -31,17 +31,73 @@ def process_chars_added(diff_added_segments):
...
+
+:Returns:
+    float
+
+:Example:
+    ..code_block :: python
+
+        >>> from revscoring.features import diff
+        >>> list(
+            extractor.extract(609079959, [diff.proportion_of_chars_added]))
```
halfak: Wow.  This looks wrong.  It should not return a value larger than 1.

Ladsgroup: The feature is obviously wrong [1] but I can't understand intention of that feature so I can't fix it.

[1] proportion_of_chars_*added* = chars_*removed* / \

halfak: Yikes!  Looks like that should say "chars_added / modifiers.max(revision.chars, 1)"

Ladsgroup:
Something is wrong about my repo, I fixed it and I create a new pull request

PR #108, opened 2015-05-30
Dependencies refactoring

halfak:
Refactors dependencies, implements a 'dig' function for gathering datasource roots propagates changes across the system.  For the most part, no current interfaces change.


revscoring/dependencies/dependent.py
```diff
@@ -0,0 +1,36 @@
...
+        return self.process(*args, **kwargs)
+
+    def __hash__(self):
+        return hash(('dependent', self.name))
+
+    def __eq__(self, other):
+        return hash(self) == hash(other)
+
+    def __ne__(self, other):
+        return not self == other
```
he7d3r: Just curious: why only `__eq__` uses `hash` when comparing with the `other`?

revscoring/dependencies/functions.py
```diff
@@ -0,0 +1,259 @@
...
+        return cache[dependent], cache, history
+
+    # Check if a corresponding dependent was injected into the context
+    else:
+
+        # If a dependent is in context here, replace it.
+        if dependent in context:
+            dependent = context[dependent]
+
+        # Check if the dependency is callable.  If not, we're SOL
```
he7d3r: SOL?

revscoring/dependencies/tests/test_context.py
```diff
@@ -0,0 +1,34 @@
...
+    foobar = Dependent("foobar", lambda foo, bar: foo + bar,
+                       depends_on=[foo, bar])
+    eq_(context.solve(foobar), "foobar")
+    eq_(list(context.solve([foo, bar, foobar])), ["foo", "bar", "foobar"])
+
+    # Cache context
+    context = Context(cache={bar: "baz"})
+    eq_(context.solve(foobar), "foobaz")
+
+    # Context context
```
he7d3r: Is this duplication of the word "context" on purpose? It sounds weird

revscoring/utilities/train_test.py
```diff
@@ -4,20 +4,28 @@
...
-    --label-type=<type>      Interprets the labels as the appropriate type
-                             (int, float, str, bool) [default: str]
+    -h --help               Prints this documentation
+    <scorer_model>          Classpath to an the MLScorerModel to construct and
+                            train
+    <features>              Classpath to an list of features to use when
+                            constructing the model
+    -p --parameter=<kv>     A key-value argument pair to use when constructing
+                            the scorer_model.
+    --version=<vers>        A version to associate with the model
```
he7d3r: Is this related to https://github.com/halfak/Objective-Revision-Evaluation-Service/issues/24?

he7d3r:
Looks good.

revscoring/dependencies/functions.py
halfak: Yeah... that's a euphemism that probably doesn't belong in a comment.  http://www.urbandictionary.com/define.php?term=SOL

revscoring/dependencies/tests/test_context.py
halfak: It does.  I nearly called this "ContextManager" but it seemed too verbose.  

I sort of think the conflation of terms is OK because they mean the same thing and serve the same purpose.  Once is a "call context" and the other is a "preserved context".  Thoughts?

revscoring/dependencies/dependent.py
halfak: It turns out that __ne__ is using hashing.  The == translates to a call to __eq__

he7d3r: oh, nice!

revscoring/dependencies/tests/test_context.py
he7d3r: Ok. I don't have any suggestions on this.

PR #112, opened 2015-06-07
Adds new scorer model names to standard imports in scorer_models.

halfak:


halfak:
I think we should get ride of the old model names when we do a major version release.  For now, having both is OK and will help us work.

he7d3r:
Looks good, and tests pass.

PR #114, opened 2015-06-14
Fixes Feature to not require a "processor" argument like Dependent and Datasource

halfak:


he7d3r:
Looks good and tests pass.

Issue #118, opened 2015-07-09
Can't configure balance_weights parameters from commandline


halfak:
You can't configure the balance_weights parameter for SVC when calling `revscoring train_test` because the balance_weights param is set on the train() function.

he7d3r:
Patch merged.

PR #119, opened 2015-07-09
Moves balance_labels arg to the constructor of SVC

halfak:


halfak:
Fixes https://github.com/wiki-ai/revscoring/issues/118

halfak:
You can now configure this parameter with `revscoring train_test ... -p balance_labels=false`

he7d3r:
Looks good and the tests pass.

PR #120, opened 2015-07-09
Fixes train_test table formating.

halfak:
Tables now look like this:

```
$ cat ../ores-wikimedia-config/datasets/enwiki.features_wp10.30k.tsv | ./utility train_test revscoring.scorer_models.RF ores.feature_lists.enwiki.wp10 revscoring.languages.english > foo
Accuracy: 0.5436973401619032

ROC-AUC:
-----  --------
A      0.699985
B      0.750562
C      0.783328
FA     0.90796
GA     0.849178
Start  0.860285
Stub   0.96958
-----  --------

         A     B     C    FA    GA    Start    Stub
-----  ---  ----  ----  ----  ----  -------  ------
A       21    19     4    37    50        1       0
B       47  1204   893   233   357      342      19
C       35   757  1145    51   297      546      30
FA     231   276   112  1819   807       10       0
GA     151   383   312   495  1532       41       0
Start   11   330   422     2    22     1638     449
Stub     3    44    47     4     0      416    2514
```

### DIFF
--- a/revscoring/utilities/train_test.py
+++ b/revscoring/utilities/train_test.py
@@ -115,23 +115,30 @@ def run(feature_labels, model_file, scorer_model):
     scorer_model.train(train_set)
 
     stats = scorer_model.test(test_set)
+
+    possible = list(set(actual for _, actual in stats['table'].keys()))
+    possible.sort()
+
     sys.stderr.write("Accuracy: {0}\n\n".format(stats['accuracy']))
     if 'auc' in stats['roc']:
-        sys.stderr.write("ROC-AUC: {0}\n".format(stats['roc']['auc']))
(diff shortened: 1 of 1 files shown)


ToAruShiroiNeko:
I think it would be even better if the sorting was based on class, ie instead of an alpha-sort it would be as below:

FA    A     GA    B     C    Start    Stub

I am merging this since above is more of a feature request than something problematic with the code.

halfak:
@ToAruShiroiNeko I thought about that, but currently we have no way of expression that FA > A

he7d3r:
Maybe create something similar to [TemplateData's `paramOrder`](https://www.mediawiki.org/wiki/Extension:TemplateData#Format) which defines the order of template parameters, but for the order of classes.

PR #126, opened 2015-07-16
Add some spanish bad words

marcelrf:
I used the English (regex) file as a reference (copy-paste).
So, I commented out all the code except for the word list.
Hope it helps, let me know if there's any problem or question.


halfak:
Thanks!  I've incorporated this work into https://github.com/wiki-ai/revscoring/pull/127

PR #132, opened 2015-07-17
Removes VERSION from manifest -- no longer used.

halfak:


he7d3r:
Where/how was this used before?

halfak:
It used to be used to capture the version for use in the setup.py file.  A change from a long time ago moves version information into the setup.py and the main __init__.py so that someone can do this:

```
>>> import revscoring
>>> print(revscoring.__version__)
0.4.4
```

halfak:
@he7d3r mergy merge mc-mergington?

PR #137, opened 2015-07-22
Document installation process

adamwight:


README.rst
```diff
@@ -75,8 +75,29 @@ Feature extraction:
...
+
+Packages
+---------
 In order to use this, you need to install a few packages first:
 
+You might need to install some other dependencies depending on your operating
+system.  These are for ``scipy`` and ``numpy``.  The fastest way is to install
+the stock packages,
+
+``sudo apt-get install python-numpy python-scipy``
```
yuvipanda: Perhaps mention that if you do this, you need to pass --system-site-packages when creating your virtualenv?

README.rst
```diff
@@ -75,37 +75,37 @@ Feature extraction:
...
+----------------
+If you need the Python package installer,
+
+``sudo easy_install3 pip``
 
-Linux Mint 17.1:
+Then, install this module,
 
-1. ``sudo apt-get install g++ gfortran liblapack-dev python3-dev myspell-pt myspell-fa myspell-en-au myspell-en-gb myspell-en-us myspell-en-za  myspell-fr aspell-id myspell-es``
+``sudo pip3 install revscoring``
```
halfak: 'sudo' is not a good idea here.  An install of this would not need to be system-wide -- unlike the apt-get call.

README.rst
```diff
@@ -75,37 +75,37 @@ Feature extraction:
...
 
-Ubuntu 14.04:
+You'll need to download NLTK data in order to make use of language features.
 
-1. ``sudo apt-get install g++ gfortran liblapack-dev libopenblas-dev python3-dev myspell-pt myspell-fa myspell-en-au  myspell-en-gb myspell-en-us myspell-en-za myspell-fr aspell-id myspell-es``
+``sudo python3 -m nltk.downloader omw stopwords wordnet``
```
halfak: Same concern over use of 'sudo' above.

README.rst
```diff
@@ -113,3 +113,5 @@ Authors
         * `http://halfaker.info`
     Helder:
         * `https://github.com/he7d3r`
+    Adam Roses Wight:
+	* `https://mediawiki.org/wiki/User:Adamw
```
halfak: Needs another ` char at the end.

Issue #138, opened 2015-07-22
All language utilities imported by 'revscoring.languages' (with proposal)


halfak:
This causes a lot of trouble.

1. A lot of stuff is loaded even if only one LanguageUtility is needed.  
2. All aspell and myspell packages must be installed for the system to work at all.  

So, these language utilities should be imported on demand.  This is difficult to accomplish due to the structure of the 'languages' module.  

Right now, languages/__init__.py imports all of the languages.  For example:
```
from .english import english
from .french import french
# etc.
```

I propose that we replace this pattern with a [dynamic import module](http://pythondoeswhat.blogspot.com/2011/12/easily-import-dynamically-created.html).  I haven't been able to get that one to work in my python 3.4 environment yet though -- so I expect we'll need to do a bit of research.

halfak:
Fixed by https://github.com/wiki-ai/revscoring/pull/141

adamwight:
Let's merge this awesome work (per IRC)!

Do you think it would be a benefit if we could also load languages like, `english = Language.get('en')`?  This looks like a really good application for mostly declarative configuration.  `revscoring/languages/spanish.py`, for example, is isomorphic with a simple and much more readable YAML file, like:
```
spanish:
  bad_regexes:
    - ano
    - ...
  dictionaries: enchant
  informal_regexes: ....
  utilities:
    - stem_word
    - is_badword
    - ...

halfak:
@adamwight, I don't think the yaml approach would work for anything but regexes as the functions may need custom definition.  We could have a mixture of simple yaml definitions and complex python definitions, but I'm not sure that would be any more easy it understand/maintain.  However, I do agree that there should be simple subclass of Language that does the non-declarative bits for the common cases.  E.g. RegexEnchantLanguage 

One thing I am troubled with re. Language.get() pattern is the ability to write your own language in your own module and use that as a drop-in replacement.  If you are going to use Language.get() in that circumstance, you'll need to make sure your module gets loaded first so that the language can be registered.  With the current structure of `revscoring extract_features` and `revscoring train_test` utilities, that wouldn't work (uses classpath of language utilities).  It would also break https://github.com/wiki-ai/ores' configuration strategy.  In the end, we'd be implementing dynamic loading ourselves within a python function and mimicking python's import system.  The problems I'm concerned about above are due to this incomplete mimicry.

adamwight:
[...]

PR #141, opened 2015-07-23
Selective language imports

halfak:
Language imports now only execute when explicitly imported.  This allows a developer to only install the enchant dictionaries that are to be used.


halfak:
Fixes https://github.com/wiki-ai/revscoring/issues/138

adamwight:
Looks good, and no test regressions!  Note that the tests still fail when dictionaries are missing, however.  Maybe we should have the tests report as skipped instead?

Issue #121, opened 2015-07-11
`RevisionDocumentNotFound` when scoring new pages


he7d3r:
It seems revscoring is not able to score new pages:
http://ores.wmflabs.org/scores/ptwiki/?models=reverted&revids=42835908|40200979
```javascript
{
  "40200979": {
    "reverted": {
      "error": {
        "message": "Failed to process <parent_revision.metadata>: RevisionDocumentNotFound",
        "type": "<class 'revscoring.dependencies.errors.DependencyError'>"
      }
    }
  },
  "42835908": {
    "reverted": {
      "error": {
        "message": "Failed to process <parent_revision.metadata>: RevisionDocumentNotFound",
        "type": "<class 'revscoring.dependencies.errors.DependencyError'>"
      }
    }
  }
}
```

PR #143, opened 2015-07-28
Fixes https://github.com/wiki-ai/revscoring/issues/121

halfak:
…ta_if_exists' to handle the situations when there is no parent revision or previous user revision.

Fixes https://github.com/wiki-ai/revscoring/issues/121

### DIFF
--- a/revscoring/features/tests/test_previous_user_revision.py
+++ b/revscoring/features/tests/test_previous_user_revision.py
@@ -17,3 +17,10 @@ def test_seconds_since():
         previous_user_revision.metadata: FakeRevisionMetadata(Timestamp(1))
     }
     eq_(solve(seconds_since, cache=cache), 9)
+
+    # Makes sure we don't crash when there was not previous user revision
+    cache = {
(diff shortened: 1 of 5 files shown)


adamwight:
Looks like an improvement, but I donno if the stuff consuming Datasources can handle Nones.

halfak:
E.g. https://github.com/wiki-ai/revscoring/blob/master/revscoring/datasources/parent_revision.py#L16
and https://github.com/wiki-ai/revscoring/blob/master/revscoring/datasources/diff.py#L10

revscoring/features/tests/test_previous_user_revision.py
```diff
@@ -17,3 +17,10 @@ def test_seconds_since():
         previous_user_revision.metadata: FakeRevisionMetadata(Timestamp(1))
     }
     eq_(solve(seconds_since, cache=cache), 9)
+
+    # Makes sure we don't crash when there was not previous user revision
```
he7d3r: "there was **no**"?

Issue #145, opened 2015-07-30
Travis CI is broken due to scipy install


adamwight:
The script hangs for over 10 minutes while provisioning scipy: https://travis-ci.org/wiki-ai/revscoring/builds/73266170 -- It's been borken forever, https://travis-ci.org/wiki-ai/revscoring/builds

There's a warning about "upgrade to our new infrastructure", fwiw.

Here's a workaround from 2014, looks like it's still necessary to do something like this.  https://gist.github.com/dan-blanchard/7045057

halfak:
I turned travis-ci off today since it was so bothersome.

PR #146, opened 2015-07-30
Attempt to defibrillate Travis-CI testing

adamwight:
See #145, we need to install scipy within the 10-minute timeout, or try other
workarounds.

Also documents how to test from the command line.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,13 +3,13 @@ docopt == 0.6.2
 mediawiki-utilities==0.4.14
 mwparserfromhell == 0.3.3
 nltk == 3.0.1
-nose == 1.3.4
-numpy == 1.9.2
+nose >= 1.3.4
+numpy >= 1.9.2
 pytz == 2015.4
-requests == 2.5.3
-setuptools == 15.1
+requests >= 2.5.3
+setuptools >= 15.1
 pyenchant == 1.6.6
-scipy == 0.14.0
-scikit-learn == 0.15.2
+scipy >= 0.14.0
+scikit-learn >= 0.15.2
 yamlconf == 0.0.1
 tabulate==0.7.5
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,13 @@ addons:
       - myspell-en-us
       - myspell-en-za
       - myspell-fa
+      - myspell-he
       - myspell-fr
       - myspell-pt
       - python3-dev
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install -y g++ gfortran liblapack-dev libopenblas-dev python3-dev aspell-id hunspell-vi myspell-pt myspell-fa myspell-en-au  myspell-en-gb myspell-en-us myspell-en-za myspell-es myspell-fr
+  - sudo apt-get install -y g++ gfortran liblapack-dev libopenblas-dev python3-dev aspell-id hunspell-vi myspell-pt myspell-fa myspell-en-au  myspell-en-gb myspell-en-us myspell-en-za myspell-es myspell-fr myspell-he
   # Begin crazy workaround to install precompiled py modules.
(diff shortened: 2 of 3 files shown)


adamwight:
Not quite there, yet...
https://travis-ci.org/wiki-ai/revscoring/builds/73338921

adamwight:
Tests pass in Travis... now, why aren't the silly traffic lights appearing in github?

adamwight:
Build time is down to 73 seconds... merge me!

requirements.txt
```diff
@@ -3,13 +3,13 @@ docopt == 0.6.2
...
+numpy >= 1.9.2
 pytz == 2015.4
 requests >= 2.0.0, < 3.0.0
-setuptools == 15.1
+setuptools >= 15.1
 pyenchant == 1.6.6
-scipy == 0.14.0
-scikit-learn == 0.15.2
+scipy >= 0.14.0
+scikit-learn >= 0.15.2
```
halfak: This version is important.  Our models can be broken by scipy upgrades.   The others I feel less strongly about, but @yuvipanda suggested that we ought to just specify specific version. 

Regardless, why did you choose to set some versions as ">=" and leave others as "=="?

adamwight: Good questions.  I think I can actually lock the versions now that I'm using miniconda to download and install the modules.  The reason I changed some of them to >= was to make it possible to install the available precompiled py modules, which are sometimes a few versions away.

Lemme play with the versions and see whether I can find a better compromise.

adamwight: Well, how do you feel about scipy 0.15.1 and scikit-learn 0.16.1?

I'm using miniconda because the pip download was timing out.  Apparently that's a known problem.  Here's the list of package versions available through miniconda:
http://docs.continuum.io/anaconda/pkg-docs

IMO the best solution would be to port our app to work with the newer versions of these two modules, then lock == to those versions.

halfak: scipy is locked at 0.14.0 because that's what we get from the repos with Debian Jessie

If we switch to scikit-learn 0.16.1, we'll need to re-generate our models.  That shouldn't be preventatively difficult, but I don't want to do it unnecessarily.  scikit-learn is our most important dependency to freeze.

halfak:
@yuvipanda how difficult would it be to change all of the things to reference scikit-learn 0.16.1 instead of 0.15.2

@adamwight how critical is it to use this more recent version and will we need to update to a recent version again soon or can we stick with 0.16.1 for a while?

adamwight:
I'm going to try to change my approach, and find a way to run CI without changing the Python module requirements.

Currently, scipy is the dealbreaker and not scikit-learn, so it's not worth the trouble to upgrade, rebuild the models, etc.

The issue with scipy is that installing via pip takes longer than a 30-second timeout built into Travis, so I'm using a binary package installer to get that down to a few seconds, but the versions don't match what we want.  There are a few potential workarounds, * emit download progress information from pip so we don't hit the "quiet for 30 seconds" timeout, * build and upload a scipy package for the same version we're using, * find another binary installer which makes the right version available.

halfak:
Our scipy requirements can be looser.  If we know the tests pass with a higher version, let's just update the requirements.txt to include that range.

yuvipanda:
Can't we just use apt-get to install scipy?

adamwight:
> Can't we just use apt-get to install scipy?

Hey yeah, we can do that since 3f2f77652ab!  Previously, the issue had been that there were many [different](https://packages.debian.org/search?keywords=python3-scipy) versions of the package depending on your platform.

adamwight:
D'oh!  Travis CI uses Ubuntu Precise, which is still at scipy 0.9.0: http://packages.ubuntu.com/precise/python3-scipy

halfak:
@adamwight This pr has been languishing for a while.  What does it need in order to get merged?

halfak:
Closing in favor of the PR that I just self-merged that got Travis working again.

PR #154, opened 2015-08-04
Generalized import_from_path() utility to work with both modules and attributes of modules.

halfak:


halfak:
FYI: This fix was necessitated by the selective language imports, but it is a generalization we should have had in the first place.

PR #155, opened 2015-08-04
Generalized requirements to ranges of minor versions

halfak:


yuvipanda:
This should be ok if we pray hard enough that everyone uses semantic versioning

PR #158, opened 2015-08-06
Adds RegexLanguage class and implements all current languages as RegexLanguages

halfak:


halfak:
This change fixes https://github.com/wiki-ai/revscoring/issues/159

Issue #135, opened 2015-07-17
Return an error for when wp10 can't find a revision


halfak:
A revision should be required to perform a wp10 classification

This doesn't return an error and it should: http://ores.wmflabs.org/scores/enwiki/wp10/671913437347480/

PR #165, opened 2015-08-07
Throw error when building a parse tree of a non-existent revision.

halfak:
Cleans up error handling so that an error is thrown when trying to parse a missing revision.  This fixes https://github.com/wiki-ai/revscoring/issues/135.


halfak:
Sitting for 25 days.  @Ladsgroup @ToAruShiroiNeko @he7d3r, got a minute?

revscoring/datasources/tests/test_revision.py
```diff
@@ -21,7 +22,6 @@ def test_content():
     eq_(solve(revision.content_tokens, cache=cache),
         ["This", " ", "is", " ", "a", " ", "foobar", "  ", "<td>"])
 
-
```
Ladsgroup: This part is not PEP8 compliment, before any first-order function definition should be two empty lines

PR #168, opened 2015-08-09
Improves logging and errors

halfak:
In extractors, dependencies and the diff datasource module.

### DIFF
--- a/revscoring/datasources/diff.py
+++ b/revscoring/datasources/diff.py
@@ -1,12 +1,21 @@
+import logging
+import time
+
 from deltas import segment_matcher
 from deltas.tokenizers import wikitext_split
 
 from . import parent_revision, revision
 from .datasource import Datasource
 
+logger = logging.getLogger(__name__)
 
 def process_operations(a, b):
-    return [op for op in segment_matcher.diff(a, b)], a, b
+    start = time.time()
+    operations = [op for op in segment_matcher.diff(a, b)]
+    logger.debug("diff() of {0} and {1} tokens took {2} seconds."
+                 .format(len(a), len(b), time.time() - start))
+
+    return operations, a, b
 
 operations = Datasource("diff.operations", process_operations,
                         depends_on=[parent_revision.tokens, revision.tokens])
--- a/revscoring/dependencies/dependent.py
+++ b/revscoring/dependencies/dependent.py
@@ -6,7 +6,7 @@
 import logging
 from functools import wraps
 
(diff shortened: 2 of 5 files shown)


halfak:
This is 23 days old.  @Ladsgroup @ToAruShiroiNeko @he7d3r , got a minute?  This one should be easy.

revscoring/datasources/diff.py
```diff
@@ -5,15 +8,21 @@
...
 def process_operations(parent_revision_text, revision_text):
     parent_revision_text = parent_revision_text or ''
     revision_text = revision_text or ''
 
+    start = time.time()
     a = wikitext_split.tokenize(parent_revision_text)
     b = wikitext_split.tokenize(revision_text)
+    operations = [op for op in segment_matcher.diff(a, b)]
+    logger.debug("diff() of {0} and {1} tokens took {2} seconds."
+                 .format(len(a), len(b), time.time() - start))
```
he7d3r: Why the interest in the length of `a` and `b`?

revscoring/dependencies/dependent.py
```diff
@@ -31,7 +31,8 @@ def __init__(self, name, process=None, depends_on=None,
         self.calls = 0
 
     def __call__(self, *args, **kwargs):
-        logger.debug("Executing {0}.".format(self))
+        logger.debug("Executing {0} ({1} calls so far)."
+                     .format(self, self.calls))
         self.calls += 1
```
he7d3r: Since this is already inside a call, shouldn't it be swapted with `self.calls += 1` so that it says "1 calls so far" on first run?

revscoring/datasources/diff.py
halfak: Diff detection time is substantially affected by the length of `a` and `b`

revscoring/dependencies/dependent.py
halfak: I don't see a clear benefit for either side of the increment.  Since python starts at `0`, I'm partial to also starting at `0`.

he7d3r: Looks good then.
Is it a problem that "All checks have failed" or can this be safely merged?

he7d3r:
Is it a problem that "All checks have failed" or can this be safely merged?

halfak:
Yeah.  I don't think that those checks have ever passed.  We should probably turn them off.  I'll do that asap. 

Also, it looks like a rebase is in order.

halfak:
@he7d3r {{done}}

PR #175, opened 2015-08-31
Jiggle dependencies to fit available python deb packaging

adamwight:
Note that some dependencies have been rolled back and we'll need to test for
regressions and incompatibilities.

Bug: T107972

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,14 +1,14 @@
 deltas >= 0.3.1, < 0.3.999
 docopt >= 0.6.2, < 0.6.999
 mediawiki-utilities >= 0.4.14, < 0.4.999
-mwparserfromhell >= 0.3.3, < 0.3.999
-nltk >= 3.0.1, < 3.0.999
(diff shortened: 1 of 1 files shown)


yuvipanda:
I think we should just pin them to specific versions than specify a range. Upgrading on the hosts will require us to build a new package and roll it out, so we'll always know the exact version...

halfak:
I don't want to tie 'revscoring' to our installation context so strongly.  Others should be able to use this module in their own context.

yuvipanda:
I don't see how specifying exact versions prevents other people from using
it?


On Tue, Sep 1, 2015 at 3:43 PM, Aaron Halfaker <notifications@github.com>
wrote:

> Merged #175 <https://github.com/wiki-ai/revscoring/pull/175>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/wiki-ai/revscoring/pull/175#event-397816254>.
>



-- 
Yuvi Panda T
http://yuvi.in/blog

PR #176, opened 2015-09-01
Added docopt.py module to revscoring/revscoring.

aetilley:


halfak:
It looks like you are adding the docopt module to address the 'docopt' dependency that revscoring has.  You don't need to add 'docopt' to the project though.  You can simply have it installed in your system (virtualenv) and revscoring should work.

PR #181, opened 2015-09-05
Fixes and tests error pickability.

halfak:
This was an issue when handling errors from celery in ORES.


halfak:
Should be uncontroversial.  No change in functionality.  I'm **boldly** merging.

PR #187, opened 2015-09-12
User lookup

halfak:


ToAruShiroiNeko:
Simple typo that didn't cause an error but effectively broke scores

he7d3r:
Was this part of the fix for [m:Talk:Objective Revision Evaluation Service#ORES revert models super negative](https://meta.wikimedia.org/wiki/Talk:Objective_Revision_Evaluation_Service#ORES_revert_models_super_negative)?

halfak:
Yes.  This was the primary issue.  We had some secondary issues once we fixed this.  We also decided that mwapi should be logging warnings that the API returns.  See https://github.com/mediawiki-utilities/python-mwapi/pull/20  We should have a new version of that out soon.

PR #192, opened 2015-09-19
Batch extraction

halfak:
Increases the speed of feature extraction by requesting data in batch.

### DIFF
--- a/revscoring/utilities/extract_features.py
+++ b/revscoring/utilities/extract_features.py
@@ -29,16 +29,20 @@
                                  [default: <stdout>]
         --verbose                Print logging information
 """
+import logging
 import sys
-import traceback
+from itertools import islice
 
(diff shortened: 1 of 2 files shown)


revscoring/utilities/extract_features.py
```diff
@@ -76,28 +82,46 @@ def read_rev_labels(f):
...
+        if len(batch_rev_labels) == 0:
             break
-        except RevisionNotFound:
-            sys.stderr.write("?")
-            sys.stderr.flush()
-        except Exception:
-            sys.stderr.write(traceback.format_exc() + "\n")
-
-    sys.stderr.write("\n")
+        else:
```
Ladsgroup: This line is redundant, by breaking in line 94. there is no need for this.

PR #198, opened 2015-10-08
Substantial improvements to documentation 

halfak:
* Adds simple examples to the main docs page.  
* Cleans up and fixes all possible intersphinx linking
* Simplifies example in readme.


README.rst
```diff
@@ -7,105 +7,57 @@ Wikipedia.
...
+
+You'll also need to install [enchant](https://enchant.org) compatible
+dictionaries of the languages you'd like to use.  We recommend the following:
+
+* ``languages.english``:  myspell-en-us myspell-en-gb myspell-en-au
+* ``languages.french``: myspell-fr
+* ``languages.spanish``: myspell-es
+* ``languages.vietnamese``: hunspell-vi
+* ``languages.hebrew``: myspell-he
+* ``languages.portugueses``: myspell-pt
```
he7d3r: Unwanted plural?
BTW: is there anything for dealing with both pt-PT and pt-BR? (as in en-US vs en-GB vs en-AU)

halfak: ha!  good catch.  

So, re. sub-types of pt.  It seems that the myspell-pt package will install myspell-pt-br and myspell-pt-pt, so we don't need to explicitly list those.

PR #199, opened 2015-10-17
fix examples: import mwapi, not import mw.api

nealmcb:


nealmcb:
Note, there are also occurrences of "mw.api" in examples/*.py and ipython/*.ipynb that should also be fixed, assuming this one makes sense.

halfak:
Thanks for the PR.  We're in the middle of fixing examples.  See https://github.com/wiki-ai/revscoring/pull/198.  I want to get that merged and then figure out what issues are remaining.

PR #200, opened 2015-10-17
Adds include-revid param to extract_features utility.

halfak:


halfak:
So, if we have a dataset that has a revid column in it, we can pass that dataset through `cut -f2-` and it will trim the first column off before sending it to `train_test`

nealmcb:
Excellent!  Great for providing all-in-one and join-able datasets for further exploration!

PR #201, opened 2015-10-17
Dockerfile that builds successfully

nealmcb:
I kept running into problems installing the python3-dev and hunspell-vi packages on my own Ubuntu trusty 14.04 desktop, so I decided it would be easier to install it in a clean Ubuntu environment via docker.  This builds on the ubuntu:trusty docker image.  See https://hub.docker.com/_/ubuntu/

I got a number of errors or warnings during the build and when trying to run things, so this may not be suitable for use yet, but should be a start.

Self-contained system-level tests, perhaps using sample model files, would be helpful.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,11 +2,11 @@
 # Install wikimedia runscoring, with dependencies
 # See: https://github.com/wiki-ai/revscoring
 
-# Build via docker build --rm -t nealmcb/revscoring:0.1 .
+# Build via docker build --rm -t nealmcb/revscoring:0.3 .
 
-FROM ubuntu:trusty
+FROM jupyter/notebook
 
-RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get upgrade -y
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && DEBIAN_FRONTEND=noninteractive apt-get upgrade -y
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
   python3-dev \
@@ -25,8 +25,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
   myspell-fr \
   myspell-es \
   hunspell-vi \
-  myspell-he \
-  python3-pip
+  myspell-he
 
 RUN pip3 install --user revscoring
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+# Dockerfile for building a Docker container.  See https://www.docker.com/
+# Install wikimedia runscoring, with dependencies
+# See: https://github.com/wiki-ai/revscoring
+
+# Build via docker build --rm -t nealmcb/revscoring:0.1 .
+
+FROM ubuntu:trusty
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get upgrade -y
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
+  python3-dev \
+  python3-numpy \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+  myspell-pt \
+  myspell-fa \
+  myspell-en-au \
+  myspell-en-gb \
+  myspell-en-us \
+  myspell-en-za \
+  myspell-fr \
+  myspell-es \
+  hunspell-vi \
+  myspell-he \
```
halfak: You'll want aspell-id too.  this was regretfully left out of the README, but it is needed for tests to pass.

nealmcb:
Besides adding the aspell-id, this also now builds on jupyter/notebook (which in turn is based on Ubuntu:trusty).  This means that running it via e.g.

    sudo docker run -it -p 8888:8888 -v localnotebooks:/notebooks nealmcb/revscoring:latest

will allow you to interact via a Jupyter notebook, loading and saving code/results/data from a local folder.

halfak:
So, I just built and started up a docker image from this docker file.  Please forgive me, but I'm still new to docker.  

It looks like this will allow a user to have an encapsulated environment for ipython notebook with revscoring installed and updated.  Is that right?  It seems like it would be difficult to use this for development and it wouldn't really help someone make use of `revscoring` as a dependency of another project (e.g. https://github.com/wiki-ai/ores).  Note that, for `ores` development, we provide a Vagrantfile.  It might make sense to also have an option to use Docker there.  

So, I'm not sure what this docker file gets us.  I'm sure I'm missing something.

yuvipanda:
Woo awesome!

Have you considered using a VOLUME command or similar to mount the source repo onto the container? That'll allow more rapid development, since you wouldn't have to rebuild the docker container after every change...

Also remember that the packages you get here will be different from the versions we use in WMF production, since that's based on debian jessie. This isn't really a 'bad thing' as much as a thing to keep in mind when debugging any issues :)

halfak:
If @yuvipanda grocks the value, that's good enough for me.  @nealmcb, I'll wait for you to comment re. VOLUME before I merge.

nealmcb:
Thanks @yuvipanda.  Since you can use the -v option from the command line to share anything you like, as I do above to share notebooks from the host to the container, I think we're covered for now.  I expect we'll want more stuff along the line, but if this works for you as is, let's go with it for now.

halfak:
Thank you!

nealmcb:
@halfak I missed your comment about usefulness.  I see the main use cases as:

* Very easy deployment for newbies
* clean testing environment for development
* potential for automated system testing for commits - want at least some simple system test code
* Could adapt / change Dockerfile to easily test on different environments: Debian, Ubuntu, Redhat, openSUSE, etc.

Docker is not intended to help someone make use of revscoring as a dependency of another project which just wants to embed the code in a larger body of code.  Pypi etc. remain great for that.  For ORES in particular, I'm guessing it could indeed be useful to deliver ORES as a docker image, using a different but perhaps related Dockerfile.  I haven't looked at ORES, but perhaps it could build "FROM" this Dockerfile and add its own server dependencies and startup code.

We might also want to have a Dockerfile that includes all the latest code under wiki-ai, and/or perhaps also have Dockerfiles for the other individual parts of wiki-ai.  We'll see where this leads, and how to structure them and reuse configuration stuff.  So this is just a start, and we'll see where it leads.

halfak:
Thanks for the explanation @nealmcb.  I'm not sure I see how someone would use this dockerfile/image/instance for development/testing.  I see how it is good for doing system tests though.

PR #206, opened 2015-10-30
Adds Italian, German and Dutch with tests & docs

halfak:


revscoring/languages/dutch.py
```diff
@@ -0,0 +1,177 @@
...
+    r"dombo",
+    r"domme",
+    r"eigelijk",
+    r"godverdomme",
+    r"groetjes",
+    r"gwn",
+    r"hoi",
+    r"hal+o+",
+    r"heb",
+    r"hee+[jyl]",
```
Krinkle: The word `heel` (meaning: whole or very, e.g. whole year or very large) seems too common to include here. Not particularly informal.

revscoring/languages/dutch.py
```diff
@@ -0,0 +1,177 @@
...
+    r"houd?",
+    r"(hoi+)+",
+    r"hoor",
+    r"izan",
```
Krinkle: I've never heard this word and can't seem to find any use of it in any Dutch material (except for a few rare uses of it as part of a name of something that isn't Dutch). It also doesn't sound or look like any existing word.

Ladsgroup: It's backward of Nazi. My friend who was checking the list, realized it

PR #210, opened 2015-11-07
No custom booleans work, so this implements and_(), and or_() instead.

halfak:


halfak:
Turns out that `__and__` and `__or__` only support bitwise operations.  Lesson learned.  Surprisingly the old test cases passed anyway.  It looks like python's weird boolean operations prevented me from knowing that my custom weird boolean operations weren't working.  Fun confusion there.

PR #215, opened 2015-11-22
Add global user age as a feature

Ladsgroup:


revscoring/datasources/types.py
```diff
@@ -66,3 +66,9 @@ def __init__(self, id=None, name=None, editcount=None, registration=None,
             if block_reason is not None else None
         self.block_expiry = str(block_expiry) \
             if block_expiry is not None else None
+
+        if globalinfo is None:
```
halfak: We should probably store globalinfo.  That way, other features can make use of it.  We really shouldn't be figuring out "globalage" within the datasource itself.  I think the datasource should just provide the raw data and features can figure out how they'll make use of that data.

revscoring/extractors/api.py
```diff
@@ -239,6 +239,12 @@ def query_users_by_text(self, user_texts, batch=50, **params):
                                        ususers=batch_texts, **params)
 
                 for user_doc in doc['query'].get('users', []):
```
halfak: Why not make global_user_info it's own thing?  Maybe having it be a subquery within user_info is unnecessary.

halfak:
@Ladsgroup, it looks like this branch got overwritten.  Is that intentional?

Ladsgroup:
No, I actually made a mess...

halfak:
I'm going to close this since it's just clutter at the moment.

PR #228, opened 2015-12-13
Add variations of existing Portuguese badwords

he7d3r:
Based on current [Salebot's list](https://pt.wikipedia.org/w/index.php?oldid=44188785).


halfak:
Awesome!  Thank you for expanding the tests too. :)

PR #233, opened 2015-12-29
Major refactor.  Datasources tree structure.  Everything is a feature set.  

halfak:
Introduces datasources.meta & features.meta -- collections of meta operators that perform common operations on `revscoring.Datasource`.

Refactors the structure of datasources to resemble a `revision_oriented` tree:

* revision
  * page
    * namespace
    * creation
  * parent
    * user
    * page
      * namespace
  * user
    * info
    * last_revision
      * page
        * namespace
 * diff

Features is now a commons space for collections of features that map against this datasource tree and discover it's shape through inspection.  

* `features.bytes` -- Features of the number of bytes of content, byte length of characters, etc.
* temporal -- Features of the time between events of a interest.  E.g. `revision.user.last_revision.seconds_since`
* `features.wikibase` -- Features of wikibase items and changes made to them.  E.g. `revision.diff.property_changed('P31')`
* `features.wikitext` -- Features of wikitext content and differences between revisions.  E.g. `revision.diff.uppercase_words_added`
* `features.revision_oriented` -- Basic features of revisions.  E.g. `revision.user.text_matches(r'.*Bot')`

Languages now also contain collections of features.  So far, we have 4 language feature collection types implements:

* `stemmed` -- Features of stemmed content.  Requires `stem_word()`.  E.g. `revision.diff.stem_prop_delta_increase`
* `dictionary` -- Feature of dictionary word matches and non-matches.  E.g. `revision.diff.dict_words_added`
* `stopwords` -- Features of stopwords and non-stopwords. E.g. `revision.diff.non_stopword_prop_delta_sum`
* `regex_matched` -- Features of regex matches applied revisions and changes.  This is useful for matching "badwords" and "informals" in free text.  E.g. `revision.diff.matches_added`

Finally, `revision.Extractor` also takes on a tree-like structure and introspection strategy for matching it's own tree of "context" (`revision.Datasource`s that use the API/DB/XML to prime the dependency tree).  

Tests are added all across the board.  This brings out test coverage from 81% to 91%, but there's a bit more to do, so it might go even higher.


revscoring/features/wikibase/tests/alan_touring.json
```diff
@@ -0,0 +1 @@
...P646", "snaktype": "value"}, "type": "statement", "references": [{"snaks-order": ["P248", "P577"], "snaks": {"P248": [{"hash": "62e345446a0c505d83139055fc6affea7d82887b", "datavalue": {"value": {"numeric-id": 15241312, "entity-type": "item"}, "type": "wikibase-entityid"}, "property": "P248", "snaktype": "value"}], "P577": [{"hash": "326dbe1cf6754fbdae762db3874856de204141f4", "datavalue": {"value": {"before": 0, "timezone": 0, "calendarmodel": "http://www.wikidata.org/entity/Q1985727", "time": "+2013-10-28T00:00:00Z", "after": 0, "precision": 11}, "type": "time"}, "property": "P577", "snaktype": "value"}]}, "hash": "af38848ab5d9d9325cffd93a5ec656cc6ca889ed"}], "id": "Q7251$FBBEEFEF-1BD3-4625-966C-94C44D375964"}], "P1343": [{"qualifiers": {"P854": [{"hash": "3dbfcd60b0730da8cb7240e581a90685b9e3b8df", "datavalue": {"value": "http://www.larousse.fr/encyclopedie/personnage/Alan_Mathison_Turing/147690", "type": "string"}, "property": "P854", "snaktype": "value"}]}, "type": "statement", "id": "Q7251$4D7F2655-DA96-46A1-BAD2-0A9CDB66F94B", "rank": "normal", "qualifiers-order": ["P854"], "mainsnak": {"hash": "3c5601ff703aba852dc449f7fa2e7bd940e53a80", "datavalue": {"value": {"numeric-id": 17
```
Ladsgroup: I'm not what this sign means, We should check if it's valid json file

halfak: It's hard to tell what you are referring to, but the test cases pass, so I suspect that the JSON is valid.

revscoring/features/wikitext/datasources/edit.py
```diff
@@ -0,0 +1,312 @@
...
+from deltas import segment_matcher
+
+from ....datasources import Datasource
+from ....datasources.meta import filters
+from .tokenized import TokenIsInTypes, is_uppercase_word
+
+logger = logging.getLogger(__name__)
+
+
+class Diff:
```
Ladsgroup: it's against pylint. We should use class Diff(object):

revscoring/features/wikitext/datasources/parsed.py
```diff
@@ -0,0 +1,239 @@
+import re
+
+import mwparserfromhell
+
+from ....datasources import Datasource
+from ....datasources.meta import filters, mappers
+
+
+class Revision:
```
Ladsgroup: same here

revscoring/features/wikitext/datasources/parsed.py
```diff
@@ -0,0 +1,239 @@
...
+
+        if name is None:
+            name = "{0}({1})" \
+                   .format(self._name + ".template_names_matching",
+                           regex.pattern)
+
+        return filters.regex_matching(regex, self.template_names, name=name)
+
+
+def _process_wikicode(text):
```
Ladsgroup: These stuff (until line 218) I think it's better to be used as static methods in Revision class, we can use it without instantiating the REvision class, like:
from datasources import parsed
link = parsed.Revision._extract_external_link_url(foo)

revscoring/features/wikitext/datasources/tokenized.py
```diff
@@ -0,0 +1,437 @@
+import re
+
+from deltas import wikitext_split
+
+from ....datasources import Datasource
+from ....datasources.meta import filters, frequencies, mappers
+
+
+class Revision:
```
Ladsgroup: same here

revscoring/features/wikitext/datasources/tokenized.py
```diff
@@ -0,0 +1,437 @@
...
+class TokenIsInTypes:
+
+    def __init__(self, types):
+        self.types = set(types)
+
+    def filter(self, token):
+        return token.type in self.types
+
+
+def _process_tokens(text):
```
Ladsgroup: We can puut this one into diff class as static method

revscoring/features/wikitext/tests/test_chars.py
```diff
@@ -0,0 +1,228 @@
...
+from .. import revision
+from ....datasources import revision_oriented
+from ....dependencies import solve
+
+r_text = revision_oriented.revision.text
+p_text = revision_oriented.revision.parent.text
+
+
+def test_chars():
+    cache = {p_text: "This is some nice text.",
```
Ladsgroup: :)))))

revscoring/features/wikitext/datasources/parsed.py
halfak: Regretfully, pickle will fail when I include these as static methods on the class.

revscoring/features/wikitext/datasources/tokenized.py
halfak: What is the "same here"?

revscoring/features/wikitext/datasources/edit.py
halfak: Woops.  It looks like this file shouldn't be here anymore.

halfak:
I'm getting 0 qualifiers added/removed between the two versions of Alan Touring's item.  This seems wrong.  Can you have a look @Ladsgroup ?

revscoring/features/wikitext/datasources/edit.py
halfak: As discussed in chat, this is only necessary python 2.x.  In python 3, inheritance from `object` is implied.

revscoring/features/wikitext/datasources/parsed.py
halfak: As discussed in chat, this is only necessary python 2.x.  In python 3, inheritance from `object` is implied.

revscoring/features/wikitext/datasources/tokenized.py
halfak: As discussed in chat, this is only necessary python 2.x.  In python 3, inheritance from `object` is implied.

revscoring/features/wikitext/datasources/tokenized.py
halfak: This has the same pickling issue we discussed earlier.  If we want static methods, we could implement something like this: http://stackoverflow.com/questions/1914261/pickling-a-staticmethod-in-python

halfak: I looked into this and it seems messy at best.  I think it's more simplistic to leave it as a module-level protected function.

halfak:
OK.  Just finished documentation.  Last thing to do before this gets merged is to finish updates to editquality and wikiclass.

Ladsgroup:
\o/

PR #238, opened 2016-02-04
Add test as stand-alone utility

halfak:


revscoring/utilities/train_test.py
```diff
@@ -127,8 +127,10 @@ def _train_test(scorer_model, observations, test_prop):
 
     logger.info("Training model...")
     scorer_model.train(train_set)
+    scorer_model.train_set = train_set
```
halfak: Would likely take up a lot of space to store this and the test set.  I'm not sure I understand why we'd need to store either the training or test set.

Ladsgroup: Thanks for the input, I thought about it too, and it's my great concern right now. Do you think other attributes of scorer_model such as "stats" or "test_statistics" are enough or we should attach all values of the test to it? Also, we don't need to store the train set at all probably.
My reason to store is to make it pickle-able so we can load it later and use it in the stand-alone script for example. If you think another approach is better, please do tell and I'll implement it.

halfak:
Here's how I imagine using this utility.  

## Step 1: Train the model on the training set
Here, I'm using the first 75% of the 20k observations to train the model.
```
$ cat datasets/wikidatawiki.features_damaging.tsv | \
  head -n 15000 | \
  revscoring train ... > \
  models/wikidatawiki.damaging.pretest.model
```

## Step 2: Test the model on the test set
Here, I'm using the last 25% of the 20k observations to test the model and store the test statistics with the model file. 
```
$ cat datasets/wikidatawiki.features_damaging.tsv | \
  tail -n 5000 | \
  revscoring test
    models/wikidatawiki.damaging.pretest.model
    -s pr -s roc ... > \
  models/wikidatawiki.damaging.model
```

We might then delete the "pretest" model file.  If we want, we can keep the `train_test` utility for making training and testing in the same step easy since this is mostly what we want for practical work around ORES.

halfak:
Just thinking that we would likely want to standardize the train and test set.  This is a set of commands that would put them into separate files.  Note that this assumes that the features_damaging file has already been randomly shuffled.

```
$ cat datasets/wikidatawiki.features_damaging.tsv | \
  head -n 15000 > datasets/wikidatawiki.features_damaging.train_15k.tsv
$ cat datasets/wikidatawiki.features_damaging.tsv | \
  tail -n 5000 > datasets/wikidatawiki.features_damaging.test_5k.tsv
```

Ladsgroup:
I haven't test it properly but I think it's now what you want :)

revscoring/utilities/test.py
```diff
@@ -0,0 +1,116 @@
...
+                   [--version=<vers>]
+                   [--values-labels=<path>]
+                   [--model-file=<path>]
+                   [--label-type=<type>]
+                   [--scale]
+                   [--debug]
+
+    Options:
+        -h --help               Prints this documentation
+        <scorer_model>          Classpath to an the MLScorerModel to construct
```
halfak: This should be a path to a model file that has already been trained -- not the scorer model class.

revscoring/utilities/test.py
```diff
@@ -0,0 +1,116 @@
...
+                                and train
+        <features>              Classpath to an list of features to use when
+                                constructing the model
+        -p --parameter=<kv>     A key-value argument pair to use when
```
halfak: When a model is already trained, params aren't used.  So these can be dropped.

revscoring/utilities/test.py
```diff
@@ -0,0 +1,116 @@
...
+    model_kwargs = {}
+    for parameter in args['--parameter']:
+        key, value = parameter.split("=")
+        model_kwargs[key] = json.loads(value)
+
+    test_statistics = []
+    for stat_str in args['--statistic']:
+        test_statistics.append(TestStatistic.from_stat_str(stat_str))
+
+    scorer_model = ScorerModel(
```
halfak: c.f. the [MLScorerModel.load](https://github.com/wiki-ai/revscoring/blob/master/revscoring/utilities/score.py#L32) line from the `score` utility for loading a model from a file.

revscoring/utilities/test.py
```diff
@@ -0,0 +1,116 @@
+"""
+``revscoring test -h``
+::
+
+    Tests a scorer model.  This utility expects to get a file of
+    tab-separated feature values and labels from which to test a model.
+
+    Usage:
+        train_test -h | --help
+        train_test <scorer_model> <features> [-p=<kv>]... [-s=<kv>]...
```
halfak: train_test --> test

halfak:
Just ran a final test on ores-compute by making sure that the train_test utility still works as intended.  Merging...

PR #240, opened 2016-02-07
Adds formatted error message to the CaughtDependencyError error.

halfak:


halfak:
This is a minor issue and only affects the usability of error messages, so I'm inclined to self-merge.  I'll wait a few hours to see if anyone wants to chime in first.

PR #243, opened 2016-02-21
Generalized test statistics and JSON formatting

halfak:
All test statistics can now be applied in a flexibly way.  All test statistics have a raw JSON-able output, a formatted JSON output and a formatted string output.  Complete with tests.


Ladsgroup:
travis is not happy. Can you check why?

PR #251, opened 2016-03-17
s/caches/cache/

arlolra:
/cc @halfak


arlolra:
Abandoning in favour of #252

PR #258, opened 2016-04-02
Adding missing commas and additional entries

harej:


halfak:
Thanks for your help!

PR #259, opened 2016-04-03
Adds russian language utility

halfak:


halfak:
@putnik, what do you think?

PR #261, opened 2016-04-08
Hindi

Ladsgroup:


Ladsgroup:
Travis is not happy because of https://github.com/ContinuumIO/anaconda-issues/issues/686 
I made some changes, I hope it works

halfak:
OK.  So I figured out some things about what is going on here. 

1. The `aspell-hi` package the comes with Ubuntu 14.04 and 13.04 (what travis uses) is broken
2. An updated `apell-hi` package is under review.  See https://launchpad.net/ubuntu/vivid/amd64/aspell-hi/0.02-7
3. You can manually install this proposed package in order to get 'hi' to work in enchant from the above link, but that won't work for Travis.  

I've filed a bug against Ubuntu and recommended that they update the selected page for apt.  See https://bugs.launchpad.net/ubuntu/+source/aspell-hi/+bug/1568313

halfak:
One more thing.  Our production environment (Jessie 8.3) uses the updates 0.02-7 package 

```
$ apt-cache policy aspell-hi
aspell-hi:
  Installed: 0.02-7
  Candidate: 0.02-7
  Version table:
 *** 0.02-7 0
        500 http://http.debian.net/debian/ jessie/main amd64 Packages
        100 /var/lib/dpkg/status
```

halfak:
OK.  I learned some more about hindi.  There are apparently "signs" that are not matched by regex's `\w` class.  I did a quick test and I got the following to match whole hindi words that include signs. 

```
re.compile("[\w\u0901-\u0965]+").match("मादरचोद मादरचोद")
<_sre.SRE_Match object; span=(0, 7), match='मादरचोद'>
```

We'll want to get this implemented in https://github.com/halfak/deltas

Ladsgroup:
Sings are actually pretty common in Semitic scripts as well to show vowels, like Hindi, you need to add them for Arabic, Persian and Hebrew (each one has its own signs)

halfak:
OK!  So I've learned that the `\b` word boundary character is weird for Hindi because the signing chars, so we need a more robust way to think about word boundaries.  Here's my python terminal session that includes my experiments.

```
>>> import re
>>> re.compile(r'\b(foo|मग्गी)\b').match("foo")
<_sre.SRE_Match object; span=(0, 3), match='foo'>
>>> re.compile(r'\b(foo|मग्गी)\b').match("मग्गी")
>>> re.compile(r'\b(foo|मग्ग)\b').match("मग्गी")
<_sre.SRE_Match object; span=(0, 4), match='मग्ग'>
>>> re.compile(r'\bमग्गी\b').match("मग्गी")
>>> re.compile(r'मग्गी').match("मग्गी")
<_sre.SRE_Match object; span=(0, 5), match='मग्गी'>
>>> re.compile(r'\bमग्गी').match("मग्गी")
<_sre.SRE_Match object; span=(0, 5), match='मग्गी'>
>>> re.compile(r'(^|[\b\s])मग्गी($|[\b\s])').match("मग्गी")
<_sre.SRE_Match object; span=(0, 5), match='मग्गी'>
```

So, I think the solution is to extend the way that we match word boundaries.  Stay tuned.

halfak:
BAM!  Got it in the last commit.  We have to match word boundaries in a really weird way with hindu because the sign characters match word boundaries `\b`.  So, instead, we match non-word chars at the end of words and we'll have to be a little careful about how we write up our hindi regexes.

halfak:
So, I'd like to get this merged and *then* consider fixing our CI environment.  We can always make due with running tests locally while the CI is in a bad state.

codez266:
I also played around with hindi regexes and since '\b' will only actually match boundaries when ending with word characters, the modifiers ([u0900-u0903] and [u093A-u0957]) being non word characters do not match. Note that each such modifier in a hindi script is appended at the end of a hindi word character so the match with '\b' at the end failed.
Related [thread](https://mail.python.org/pipermail/python-list/2008-November/512205.html)

PR #263, opened 2016-04-12
Faster score utility and fixes to extractor

halfak:


revscoring/dependencies/functions.py
```diff
@@ -245,7 +245,6 @@ def _solve(dependent, context, cache, history=None, profile=None):
                 duration = time.time() - start
                 if profile is not None:
                     if dependent in profile:
-                        print("here")
```
Ladsgroup: was it me? I do this kind of crazy stuff all the time.

revscoring/extractors/api/extractor.py
```diff
@@ -71,24 +75,29 @@ def extract(self, rev_ids, dependents, context=None, caches=None,
...
         else:
             rev_id = rev_ids
-            cache = caches
+            cache = cache or {}
+            cache.update((caches or {}).get(rev_id, {}))
             return self._extract(rev_id, dependents, context, cache,
                                  profile)
 
-    def _extract_many(self, rev_ids, dependents, context, caches, profile):
+    def _extract_many(self, rev_ids, dependents, context, caches, cache,
```
Ladsgroup: this still gets cache and not ex_cache, is it intentional?

halfak: That's right.  The external interface takes a "cache", but internally, we need to do a shallow copy and make a bunch of changes so it is changed to "ex_cache" to avoid naming confusion

revscoring/dependencies/functions.py
halfak: I think it was me, actually ^_^

PR #279, opened 2016-06-07
Prepends local dir on path for utilities that use classpaths

halfak:


halfak:
This allows you to reference a classpath without having it installed.  It will search the local directory for the base-path.  This is actually expected behavior in general (since the local directory will be searched in the repl, but not by importlib.import_module).

PR #287, opened 2016-09-02
Support for sparse feature vectors 

halfak:
* Implements `FeatureVector`
* Implements meta-`Datasource`s for generating bag-of-words vectors
  * hashing (fast, portable hashing with murmur3)
  * gramming (ngrams and skipgrams)
  * selectors (TFiDF and raw key-based)
* Refactors utilities to use JSON and base85 encoded pickle blobs for `cache`.  See https://phabricator.wikimedia.org/T132580 for painful deliberations about this strategy


revscoring/utilities/extract.py
```diff
@@ -0,0 +1,283 @@
+"""
```
Ladsgroup: Why renaming this utility?

halfak:
Here's my gramming/hashing speed experiment https://gist.github.com/halfak/565d1c2153da57c5c6600cb175f20236

revscoring/utilities/extract.py
halfak: It does something differently and is no longer specific to "features"

PR #290, opened 2016-09-26
Fixes issues with tune utility

halfak:


Ladsgroup:
It would be great of the sake of documentation. Write what is wrong and what you did to fix it. Oterwise LGTM

halfak:
Ahh Gotcha.  I actually discovered these bugs while simply trying to use it.  these were regressions introduced in the last major refactoring of the utilities.  

One bug of note though is that the following code works as expected in python 3.5:
```python
Python 3.5.1+ (default, Mar 30 2016, 22:46:26) 
[GCC 5.3.1 20160330] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> def print_kwargs(**kwargs):
...     print(kwargs)
... 
>>> kwargs_a = {'foo': 1}
>>> kwargs_b = {'bar': 2}
>>> 
>>> print_kwargs(**kwargs_a, **kwargs_b)
{'foo': 1, 'bar': 2}
```

But it doesn't work in python 3.4:
```python
Python 3.4.2 (default, Oct  8 2014, 10:45:20) 
[GCC 4.9.1] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> def print_kwargs(**kwargs):
...     print(kwargs)
... 
>>> kwargs_a = {'foo': 1}
>>> kwargs_b = {'bar': 2}
>>> 
>>> print_kwargs(**kwargs_a, **kwargs_b)
  File "<stdin>", line 1
    print_kwargs(**kwargs_a, **kwargs_b)
                           ^
SyntaxError: invalid syntax
```

PR #309, opened 2017-04-29
Add Bengali language assets

Ladsgroup:
Tests fail for strange reasons that is beyond my knowledge

Bug: https://phabricator.wikimedia.org/T162620


Ladsgroup:
Also, needs https://github.com/travis-ci/apt-package-whitelist/pull/4107/files to be merged first

halfak:
We might need to disable that test.  It could be that the package is not available in our test vms.  See had to do the same for hindi and tamil.  See https://github.com/wiki-ai/revscoring/blob/master/.travis.yml#L51

Ladsgroup:
Done

revscoring/languages/tests/util.py
```diff
@@ -6,6 +6,7 @@ def compare_extraction(extractor, examples, counter_examples,
 
     for example in examples:
         wrapped = lwrap + example + rwrap
+        print(extractor.process(wrapped))
```
halfak: We should probably drop this line.

halfak:
Needs an update to requirements.txt for deltas 0.4.6

PR #317, opened 2017-05-21
Greek language assets

OdysseasKr:
Added files for Greek language. This includes the needed words from [Research:Revision scoring as a service/Word lists/el](https://meta.wikimedia.org/wiki/Research:Revision_scoring_as_a_service/Word_lists/el) and some extra ones.


Ladsgroup:
Thanks for doing this

PR #324, opened 2017-06-12
add another german badword

Pix1234:


halfak:
Please add it in alphabetical order and ad a test too. :)

PR #326, opened 2017-06-23
Fix pathological backtracking regexp

subbuss:
This particular pattern (..+)* has been responsible for several
instances of non-terminating regexps in Parsoid.


halfak:
Thank you!

PR #327, opened 2017-06-24
Tweak one regexp + update for portuguese.py

subbuss:
Not sure if you can pluck just that last commit or not. If not, I can sync up and push tomorrow.


subbuss:
Updated PR after syncing with upstream

PR #330, opened 2017-06-28
Add missing options to where docopt can find them

adamwight:
See https://phabricator.wikimedia.org/T169157


halfak:
Eek!  Good catch.

PR #335, opened 2017-07-10
Add Albanian Language Assets

codez266:
Incorporating list from https://www.mediawiki.org/wiki/Research:Revision_scoring_as_a_service/Word_lists/sq and adding Albanian language assets and associated tests


revscoring/languages/tests/test_albanian.py
```diff
@@ -0,0 +1,88 @@
...
+    r"banya",
+    r"baseballe",
+    r"basketball",
+    r"basketballnationalmannschaft",
+    r"basketbalteam",
+    r"bastard",
+    r"bator",
+    r"baturina",
+    r"beker",
+    r"ber",
```
halfak: Seems like this list has run short.

revscoring/languages/tests/test_albanian.py
```diff
@@ -0,0 +1,88 @@
...
+    r"awesome",
+    r"btw",
+    r"category",
+    r"cdo",
+    r"cool",
+    r"dreq",
+    r"eshte",
+    r"ftw",
+    r"haha",
+    r"hahah",
```
halfak: This list is cut off too

PR #338, opened 2017-07-27
Data utils

adamwight:


revscoring/utilities/tests/test_union_merge_observations.py
```diff
@@ -13,14 +14,19 @@ def test_union():
...
 
-    # Receive the output as a string buffer.
-    out_file = io.StringIO()
+    # Receive the output in a temporary file.
+    (fh, out_file) = tempfile.mkstemp()
 
     # Do the union.
-    DataUnion().union(in_files, out_file)
+    argv = in_files + ["--output", out_file]
+    revscoring.utilities.union_merge_observations.main(argv)
```
halfak: This would be a great place to put the run() function we talked about.

revscoring/utilities/union_merge_observations.py
```diff
@@ -40,36 +41,21 @@ def main(argv = None):
...
         out_file = open(args['--output'], "w")
 
-    transformation = DataUnion(id_column = args['--id-column'])
-    transformation.union(args['<input>'], out_file)
+    observations = []
+    for path in args['<input>']:
+        with open(path, "r") as in_file:
+            observations.extend(read_observations(in_file))
 
+    union_merge(observations, out_file, id_column = args['--id-column'])
```
halfak: I'm not sure this will work outside of the `with ...` block because it should call `__exit__()` on the files as soon as the block is exited.

revscoring/utilities/union_merge_observations.py
```diff
@@ -40,36 +41,21 @@ def main(argv = None):
...
 
-class DataUnion(object):
 
-    # FIXME: This doubled default is annoying.
-    def __init__(self, id_column = "rev_id"):
-        self.id_column = id_column
+def union_merge(observations, out_file, id_column):
```
halfak: Including the out_file here seems problematic.  It seems that we should have a function for doing the in-memory operations and then we can separately decide to write output to a file.

revscoring/utilities/tests/test_union_merge_observations.py
halfak: Would look something like 

`run(observations, out_file, key)`

adamwight: There's so much logic in main() that I'd rather test that.  I'm especially wary of the big read_observations mush.

revscoring/utilities/union_merge_observations.py
adamwight: out_file doesn't use a context manager :(

revscoring/utilities/union_merge_observations.py
adamwight: True, it's asymmetric in a bad way.  Fixing.

revscoring/utilities/union_merge_observations.py
halfak: `observations` is a list of generators that have yet to be read.  The files would still need to be open outside of the context block in order to be able to read them.

revscoring/utilities/union_merge_observations.py
```diff
@@ -0,0 +1,62 @@
...
+
+    if args['--output'] == "<stdout>":
+        out_file = sys.stdout
+    else:
+        out_file = open(args['--output'], "w")
+
+    observations = []
+    for path in args['<input>']:
+        with open(path, "r") as in_file:
+            observations.extend(read_observations(in_file))
```
halfak: This will load all observations into memory -- which may be substantially more than necessary depending on how many observations will be eventually merged.

revscoring/utilities/tests/test_union_merge_observations.py
```diff
@@ -0,0 +1,48 @@
...
+def test_union():
+    # Get test fixtures.
+    data_dir = os.path.dirname(__file__) + "/data"
+    in_files = [
+        data_dir + "/labeled_revisions.json",
+        data_dir + "/labeled_foo.json",
+    ]
+
+    # Receive the output in a temporary file.
+    (fh, out_file) = tempfile.mkstemp()
```
halfak: Why not use a text IO for this?

adamwight: An io.StringIO?  I wanted to test the stuff in main() as well as union_merge, so not sure how I would do this.  I need to pass main() a string that gets passed to open() as a pathname.

revscoring/utilities/union_merge_observations.py
adamwight: Yeah, it's unfortunate, and the entire output gets read into memory as well.  I made a note about this inefficiency in the commit message and the module's docstring, to do it later.  Might as well put a quick note here as well, since people will run into it and wonder as you've demonstrated.

Do you think we have to worry about it now, though?  I was going to wait until we run into the wall.  Estimating, editquality labeled revisions files are 180 bytes per row or 180MB per 1M observations.  A few of those files, Python overhead, and the worst case scenario of no duplicates would nary add up to a gigabyte.  If we start talking feature cache... but otherwise nah let's kick this down the road.

*How* to solve the problem is a cool algorithm question, though.  I don't know!  I can't think of anything clever off-hand, the tradeoff looks like disk access vs. memory.

halfak: I guess I just don't understand what the problem was with the solution I proposed.  That solution made more efficient use of memory by reading from one file at a time. 

Essentially, the strategy I proposed used `|A∪B|` memory while yours uses `|A|+|B|` memory.

revscoring/utilities/tests/test_union_merge_observations.py
halfak: I see.  Maybe it would be better to test these things separately -- *i.e.*, in units.  I guess the reason I keep pushing on this is because it's working hard read actual files when the real thing we want to test is the merge and union functionality.  This feels like an integration test.

revscoring/utilities/union_merge_observations.py
halfak: Oh!  I just had a though.  If you move the merge_union call in-context you can do something like this:

```
from itertools import chain
observations = (read_observations(open(path) for path in args['<input>'])
union_merge(chain(*observations), ...)
```

revscoring/utilities/tests/test_union_merge_observations.py
adamwight: It is an integration test :)  At some level, everything is an integration test--what we gain by writing a second test to only run union_merge() is a little granularity about where the failure is, if something goes wrong.

I would resist (to a reasonable point) taking out the integration test, because with the .extend(read_observations), a significant amount of the logic is taken out of union_merge.  That in itself is wonderful and I'm glad we did it, but I don't see why we would want to omit testing the extend?

halfak: ... don't `.extend()` at all?

revscoring/utilities/union_merge_observations.py
adamwight: > Essentially, the strategy I proposed used |A∪B| memory while yours uses |A|+|B| memory.

Awesome insight!  I hadn't even noticed that efficiency.  Remind me if I don't patch tomorrow...

TIL itertools, thanks.

adamwight: Latest patch uses generators!

PR #341, opened 2017-08-09
Revscoring 2 followup again

halfak:


revscoring/scoring/statistics/classification/tests/test_scaled_threshold_statistics.py
```diff
@@ -32,25 +32,33 @@ def test_sts():
...
+        s_rounded_threshold = round(s_stats[0], 3)
+        s_threshold_stats[s_rounded_threshold] = s_stats[1]
+
+        b_rounded_threshold = round(b_stats[0], 3)
+        b_threshold_stats[b_rounded_threshold] = b_stats[1]
+
+    count = 0
+    for threshold, s_stats in s_threshold_stats.items():
+        if threshold not in b_threshold_stats:
+            continue
```
adamwight: Why do we ignore these, rather than count?

revscoring/scoring/statistics/classification/tests/test_scaled_threshold_statistics.py
```diff
@@ -32,25 +32,33 @@ def test_sts():
...
+        b_stats = b_threshold_stats[threshold]
+        for stat_name in s_stats:
+            if s_stats[stat_name] is not None:
+                diff = (s_stats[stat_name] or 0) - (b_stats[stat_name] or 0)
```
adamwight: That's odd.  The first "or" clause only does anything if s_stats[stat_name] is literal False.

revscoring/scoring/statistics/classification/tests/test_scaled_threshold_statistics.py
```diff
@@ -32,25 +32,33 @@ def test_sts():
...
     print("Error", me)
-    assert rmse < 0.20, rmse
-
-    assert abs(balanced_sts.roc_auc() - skewed_sts.roc_auc()) < 0.10, \
-           abs(balanced_sts.roc_auc() - skewed_sts.roc_auc())
-
-    assert abs(balanced_sts.pr_auc() - skewed_sts.pr_auc()) < 0.10, \
-           abs(balanced_sts.pr_auc() - skewed_sts.pr_auc())
+    assert rmse < 0.05, rmse
+    assert count > 1000
```
adamwight: Would be nice to add some comments explaining what we're checking here.

revscoring/scoring/statistics/classification/tests/test_scaled_threshold_statistics.py
halfak: Ahh yes.  It seems that both stats are given a default value of zero even though s_stats[stat_name] couldn't be null by this point.  It also can't be false.  Seems harmless to me.

PR #357, opened 2017-09-04
Add Bosnian language assets

Ladsgroup:
Bug: T173087

### DIFF
--- a/revscoring/languages/tests/test_bosnian.py
+++ b/revscoring/languages/tests/test_bosnian.py
@@ -1,0 +1,151 @@
+import pickle
+
+from nose.tools import eq_
+
+from .. import bosnian
+from ...datasources import revision_oriented
+from ...dependencies import solve
+from .util import compare_extraction
+
+BAD = [
+    "balija",
+    "debil",
+    "debili",
+    "debilu",
+    "drolja",
+    "drolje",
+    "droljetine",
+    "droljo",
+    "fašisti",
+    "fuck",
+    "govna",
+    "govno",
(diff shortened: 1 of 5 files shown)


halfak:
Travis couldn't continue "apt-get install failed".

Can you confirm that this package exists and works on debian-stretch?

halfak:
I just saw the comment in phab that it works on Stretch.  We'll need to squash this test on travis until travis supports stretch.

revscoring/languages/tests/test_bosnian.py
```diff
@@ -0,0 +1,151 @@
...
+    element sa stabilnim izotopima koji je otkriven (renij je identificiran
+    dvije godine kasnije).
+    """
+]
+
+r_text = revision_oriented.revision.text
+
+
+def test_badwords():
+    compare_extraction(bosnian.badwords.revision.datasources.matches,
```
adamwight: I see that we use this pattern for every language, so no the fault of this revision, but I'm personally not a fan of this type of test.  We can assert that the list can be unpickled, that it's longer than 10 or something, but keeping two copies of the same list in sync seems counterproductive.

halfak: @adamwight, note that these lists diverge when someone has time to take advantage of regular expressions in the primary list.  See enwiki.py for example.  Also note that this test confirms that we are able to extract badwords when there is text surrounding it.  This is extra important when we have regular expressions involved that can match whitespace and other word breaking chars.

PR #358, opened 2017-09-11
Fixes sorting in tune utility.

halfak:


halfak:
I'll have an incoming pull request in https://github.com/wiki-ai/editquality that will update all of the tuning reports.  I got from arwiki through jawiki with this fix.

PR #359, opened 2017-09-11
Changes uk dict recommendation to aspell (from myspell)

halfak:


Ladsgroup:
This needs update in puppet I guess

PR #362, opened 2017-09-12
Natively handles bz2 compressed model files.

halfak:
This will allow us to store compressed model files and safely use them.

### DIFF
--- a/revscoring/scoring/models/model.py
+++ b/revscoring/scoring/models/model.py
@@ -7,6 +7,7 @@ All scoring models are an implementation of :class:`revscoring.Model`.
 .. autoclass:: revscoring.scoring.models.Classifier
     :members:
 """
+import bz2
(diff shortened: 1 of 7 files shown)


Ladsgroup:
We have an unhappy travis here, do flake8 :D

revscoring/scoring/models/model.py
```diff
@@ -156,7 +164,7 @@ def __init__(self, *args, scale=False, center=False, **kwargs):
         })
 
     def train(self, values_labels):
-        """
+        """models/enwiki.draft_quality.gradient_boosting.model
```
Ladsgroup: This seems like an error

halfak:
Thanks!  Got it.

halfak:
Self merging with cleaned flake8

PR #365, opened 2017-10-06
Limit memory usage of thresholds statistics

halfak:
This PR contains changes that make it possible to limit threshold statistics by rounding the "decision" value that thresholds are based on.  This allows a user to effectively reduce the number of thresholds that are stored in revscoring.Model.info for probability based models.  

https://phabricator.wikimedia.org/T177544


adamwight:
A late followup: I've smoke tested on a small training set, and the threshold rounding worked perfecftly at both 3 and 2 decimal places.

PR #366, opened 2017-10-18
Migrate from nosetests to pytests

Ladsgroup:
Bug: T174660


Ladsgroup:
This will fail both in matter of flake8 and ci tests, I'm fixing them.

Ladsgroup:
It was a fucking horrible mistake by using a too general regex, fixed now :)

revscoring/languages/tests/test_bosnian.py
```diff
@@ -1,8 +1,11 @@
 import pickle
 
-from nose.tools import eq_
+from pytest import mark
 
-from .. import bosnian
+try:
+    from .. import bosnian
+except:
+    pytestmark = mark.nottravis
```
halfak: What's going on here?

Ladsgroup: That's a great question, importing bosnian causes error in travis because the the dictionary is not there and pytestmark = mark.nottravis is a global identifier asking pytest not to touch this file. We can go with mark.skip as well. Doesn't matter.

halfak: Gotcha.  This is OK, but maybe we should have a comment there?  I'm not sure.  If this is going to be a common pattern, it shouldn't need a comment, but in this case we're only doing it twice.

Ladsgroup: Done :)

PR #376, opened 2017-12-08
(WIP) Multilabel random forest

halfak:
(@halfak made this into a PR, but it's mostly @codez266's work)


halfak:
Made this into a PR because it's easier to track this way :)

revscoring/scoring/labels.py
```diff
@@ -0,0 +1,78 @@
...
+            return label
+
+    def denormalize(self, label):
+        return label
+
+
+class Binarizer(ClassVerifier):
+
+    def __init__(self, possible_labels=None):
+        super().__init__(possible_labels)
```
halfak: This can't be None.  `possible_labels` should be a required argument

revscoring/utilities/tune.py
```diff
@@ -43,6 +44,7 @@
                                be assumed to reflect population rates.
         --minimize             If set, assume the best score is the smallest
                                value.
+        --multilabel           Whether to perform multilabel classification
```
halfak: Can't this just be added as a param in the parameter config?

revscoring/utilities/tune.py
```diff
@@ -124,6 +127,12 @@ def main(argv=None):
     if args['--scale']:
         additional_params['scale'] = args['--scale'],
 
+    additional_params['multilabel'] = False
+    if args['--multilabel']:
+        _, all_labels = zip(*value_labels)
+        labels = get_multilabel_set(all_labels)
```
halfak: Use the `--labels`, weights or rates just like in multiclass.

halfak: https://github.com/wiki-ai/revscoring/pull/376/files#diff-34e19f77339293b2504629af65db696cR117

PR #379, opened 2017-12-11
Try to fix travis

Ladsgroup:


Ladsgroup:
I think so

adamwight:
Ah—just a minute.  It doesn't look like travis ran the line that you added.

Also, we should do pip magic to force wheel installation, if that's the problem.

Ladsgroup:
It definitely did run the line that I added

adamwight:
Cool, thanks for pointing that out.  I think I skimmed over the collapsed lines.

This is a problem, though:
> Downloading scipy-1.0.0-cp35-cp35m-manylinux1_x86_64.whl

We need to restrict the scipy version to match production, ==0.17.1

On the bright side, I see that it's using the wheel, so that's taken care of!

PR #385, opened 2017-12-21
Add support for specifying label weights in multilabel classification

codez266:
Current label_weights implementation in revscoring passes **label_weights** as '**sample_weights**' during fit() of classifier. 
However, according to [docs](http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html) label weights for multilabel classification need to be passed during classifier initialization in a specific format.
This change copies the current **label_weights** argument into **class_weight** and passes it as an estimator_param when multilabel classification is performed. Additionally it deletes the **label_weights** parameter so that weights do not get passed again during fit() of estimator.

### DIFF
--- a/revscoring/utilities/cv_train.py
+++ b/revscoring/utilities/cv_train.py
@@ -99,6 +99,9 @@ def main(argv=None):
     multilabel = False
     if args['--multilabel']:
         multilabel = True
+        if label_weights:
+            estimator_params['class_weight'] = label_weights
+            label_weights = None
 
     model = ScoringModel(
         features, version=version, multilabel=multilabel,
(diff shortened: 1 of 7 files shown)


revscoring/utilities/cv_train.py
```diff
@@ -99,6 +99,9 @@ def main(argv=None):
     multilabel = False
     if args['--multilabel']:
         multilabel = True
+        if label_weights:
+            estimator_params['class_weight'] = label_weights
+            label_weights = None
```
halfak: This should be handled within the scorer model

halfak:
@codez266, I moved the label weight handling inside of sklearn.py.

codez266:
ran cv_train on wikiprojects dataset with this change, look good...

PR #388, opened 2017-12-27
Revscoring: Implement fast scoring for fast cross validation

codez266:
- Implements a new method `score_many` in sklearn.py to score a bunch of instances together to utilize numpy's underlying optimizations of matrix multiplication.
- `_cross_score` is made to call this new method to improve speeds of cross_validation. The original score method is preserved for use in ORES.
- New tests added for `score` and `score_many` as well as a test for probability classifier in test_sklearn to improve coverage.

### DIFF
--- a/revscoring/scoring/models/model.py
+++ b/revscoring/scoring/models/model.py
@@ -246,8 +246,9 @@ class Learned(Model):
         logger.debug("Training cross-validation for {0}...".format(i + 1))
         model.train(train_set)
         logger.debug("Scoring cross-validation for {0}...".format(i + 1))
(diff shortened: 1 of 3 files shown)


halfak:
How much faster is this?  It's a lot of complication to add.  I think we should quantify the benefit.

codez266:
```
RF classifier parameters used for benchmarking:
num_samples - 10k
n_estimators - 400
max_depth - 4
features - word2vec( 300 dimensions )
binary classification with True/False

Time for current cv_train - 414.65737676620483
Time for new fast cv_train - 81.77100491523743
```
Performance gain is almost 5 times.

PR #389, opened 2018-01-02
Use collection of binary classifiers in Revscoring classifier for multilabel

codez266:
- Major refactoring to eliminate sklearn's native multilabel classification and introduce per label binary classification.
- List of sklearn's binary classifiers are stored in Revscoring's classifier wrapper, one for each label.
- Tests modified.


halfak:
So, this looks overly complicated.  Why not just use the "multilabel" parameter to enable a OneVsRest classifier?  Why should there be a whole different class for OneVsRest? 

E.g.  

```
regular_gb_model = GradientBoosting(...)

multi_gb_model = GradientBoosting(..., multilable=True)
```

It seems like we should drop support for direct multilabel in the RandomForest model -- since it wasn't useful for us.

halfak:
Looks like the GB model doesn't yet work, but I think that is because we're still using the "SUPPORTS_MULTILABEL" class variable.  I think we should drop that entirely because it's no longer needed.  All classifiers support multilabel in this scenario.

PR #391, opened 2018-02-19
Add word2vec features

codez266:
- Adds a new feature for wordvectors subclassing from FeatureVector. The feature accepts word2vec parameters like name, dimesions, word limit, etc. and additionally a list of tokens datasource. 
- It returns a feature equal to the dimensions of the original vectors averaged over the tokens supplied.
- Uses a word2vec helper to load the binary file. Currently the only defined helper is the `gensim_helper` to load the vectors using the gensim library.
- This allows us to write features like:
```
import vectorizers
import wikitext
google_news_kvs = vectorizer.load_kv("google_news.bin")
revision_text_vectors = vectorizers.word2vec(
  wikitext.revision.datasources.words,
  google_news_kvs,
  name="revision.text.google_news_vectors")
```


revscoring/datasources/meta/tests/test_vectorizers.py
```diff
@@ -0,0 +1,24 @@
+from unittest.mock import patch
+import numpy as np
+
+from .. import vectorizers
+from revscoring.datasources import revision_oriented as ro
+from revscoring.dependencies import solve
+from revscoring.features import wikitext
+
+test_vectors = {'a': np.ones(200),
```
halfak: This is an order of magnitude slower than using the raw python. 

```
>>> start = time.time(); sum(sum([0]*300) for i in range(10000)); time.time() - start
0
0.09108448028564453
>>> start = time.time(); sum(sum(np.zeros(300)) for i in range(10000)); time.time() - start
0.0
0.35784435272216797
```

It would be better to use [1] * 200.   I don't mean to bring this up in a test (who cares), but rather to suggest that the whole of the submitted code should use the [0] * NUM pattern instead.

revscoring/datasources/meta/tests/test_vectorizers.py
```diff
@@ -0,0 +1,24 @@
...
+                'b': np.ones(200),
+                'c': np.ones(200)}
+
+loadvectors_full_path = vectorizers.__name__ + \
+    '.word2vec.load_kv'
+
+
+@patch(loadvectors_full_path)
```
halfak: Why is this necessary?

revscoring/datasources/meta/vectorizers.py
```diff
@@ -0,0 +1,53 @@
...
+            in the dict.
+        dim : `int`
+            The dimension of the vectors
+        limit : `int`
+            Max number of word vectors to load
+        name : `str`
+            A name for the `revscoring.FeatureVector`
+    """
+
+    def __init__(self, items_datasource, w2v_prefix="~/.word2vec/", w2v=None,
```
halfak: Should take a KeyedVectors instance as an argument.

revscoring/datasources/meta/vectorizers.py
```diff
@@ -0,0 +1,53 @@
...
+        self.dim = dim
+        self.limit = limit
+        self.w2v = self.load_kv(w2v_prefix + w2v, limit)
+        super().__init__(name, self.process, depends_on=[items_datasource])
+
+    def process(self, d):
+        return [self.w2v[w] if w in self.w2v else np.zeros(self.dim)
+                for w in d]
+
+    def load_kv(self, path, limit):
```
halfak: Should be a class method.

revscoring/features/meta/aggregators.py
```diff
@@ -10,56 +10,81 @@
...
+import numpy as np
 
 from ..feature import Feature
+from ..feature_vector import FeatureVector
 
 len_builtin = len
 sum_builtin = sum
 max_builtin = max
 min_builtin = min
+mean_builtin = np.mean
```
halfak: Should probably use `statistics.mean` instead.

test-requirements.txt
```diff
@@ -1,2 +1,3 @@
 codecov
 pytest-cov
+google-compute-engine <= 2.7.999
```
halfak: What is this used for?

codez266: problem in the boto library, workaround for that - https://github.com/RaRe-Technologies/gensim/issues/898

halfak:
I submitted a change.  Looks like tests are passing.  @codez266, please review and merge if you like my changes.  Raise concerns if you have them :)

revscoring/datasources/meta/vectorizers.py
```diff
@@ -0,0 +1,56 @@
...
+            for dir_path in ASSET_SEARCH_DIRS:
+                try:
+                    path = os.join(dir_path, filename)
+                    return KeyedVectors.load_word2vec_format(
+                        path, binary=True, limit=limit)
+                except FileNotFoundError:
+                    continue
+        else:
+            raise TypeError(
+                "load_kv() requires either 'filename' or 'path' to be set.")
```
codez266: It might be beneficial to have a test for this static method code. I'll take care of that in a followup :)

PR #392, opened 2018-03-01
Update spanish.py

MarcoAurelioWM:
added a couple of words, defined few and removed one which is not a bad word "per se".

### DIFF
--- a/revscoring/languages/spanish.py
+++ b/revscoring/languages/spanish.py
@@ -143,7 +143,7 @@ badword_regexes = [
     r"mocos",  # boogers / mucus
     r"mojon",  # shit
     r"monda",  # man part / penis
-    r"mongolo", # idiot (slang)
+    r"mongolo",  # idiot (slang)
     r"moro",  # moor / muslim (derogatory)
     r"nacio",  # ???
     r"nazi",  # nazi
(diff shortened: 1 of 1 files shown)


revscoring/languages/spanish.py
```diff
@@ -135,15 +134,16 @@
...
     r"me[- ]?castra[- ]?la[- ]?madre",  # 'I <something> your mother' ???
-    r"me[- ]?la[- ]?pelan?s?",  # "You ain't shit" ???
+    r"me[- ]?la[- ]?pelan?s?",  # I don't give a f***
     r"mi?er[dg]a+s?\w*",  # shit
     r"mimsn",  # ???
     r"ming[ao]",  # ugly person
     r"mocos",  # boogers / mucus
     r"mojon",  # shit
     r"monda",  # man part / penis
+    r"mongolo", # idiot (slang)
```
halfak: flake8 is failing here.  There should be two spaces before the comment "#"

revscoring/languages/spanish.py
```diff
@@ -183,6 +183,7 @@
     r"sida",  # AIDS
     r"sorete",  # piece of shit
     r"sudaca",  # greaser
+    r"subnormal", # retarded
```
halfak: Needs another space here too.  Then all should be good :)

revscoring/languages/spanish.py
MarcoAurelioWM: Fixing.

revscoring/languages/spanish.py
MarcoAurelioWM: Fixing those. Thanks.

halfak:
Thanks!

PR #394, opened 2018-03-08
Use global vectors for better multiprocessing and remove english_vectors

codez266:
* Experiments have shown that using global vectors in the vectorizers file works better for memory sharing - https://phabricator.wikimedia.org/T189364
* Removes the english_vectors script since we no longer need the vectors as a separate module.

### DIFF
--- a/revscoring/datasources/meta/tests/test_vectorizers.py
+++ b/revscoring/datasources/meta/tests/test_vectorizers.py
@@ -1,6 +1,7 @@
+import pickle
 from unittest.mock import patch
-import pytest
 
+import pytest
 from revscoring.datasources import revision_oriented as ro
 from revscoring.dependencies import solve
 from revscoring.features import wikitext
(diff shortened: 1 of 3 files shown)


halfak:
Cool.  Does this work for dealing with the multiprocessing/shared memory issue?

codez266:
With this, it does not OOM, but its not as fast as using the global keyword and memory usage seems to be high ~ 6GB. I'm not sure what is happening here, but we might need to revert back to using 'global'

halfak:
Weird.  That doesn't make any sense.  Imports *are* globals.  Can you demonstrate this?

codez266:
https://phabricator.wikimedia.org/T189364

codez266:
Yeah, I had a similar idea but assigning any *big* thing in self leads to explosion of memory, which I think is because doing `self.xyz=` basically means we're causing the thing to pickle.

codez266:
since it passes the speed test and looks good, going ahead to merge...

PR #399, opened 2018-05-04
Increment patchlevel to 2.2.3

adamwight:


adamwight:
Self-merging version bump

PR #400, opened 2018-05-20
Adds minimal language (just dictionary) for Basque

halfak:


revscoring/languages/basque.py
```diff
@@ -0,0 +1,16 @@
+from .features import Dictionary
+
+name = "basque"
+
+try:
+    import enchant
+    dictionary = enchant.Dict("eu")
+except enchant.errors.DictNotFoundError:
+    raise ImportError("No enchant-compatible dictionary found for 'bs'.  " +
```
kevinbazira: Should `for 'bs'` be `for 'eu'` since Basque is 'eu'?
```suggestion
    raise ImportError("No enchant-compatible dictionary found for 'eu'.  " +
```

halfak: Nice catch.  Should be fixed.

PR #403, opened 2018-06-11
Bump patchlevel to 2.2.5

adamwight:


adamwight:
Self-merging version increment

PR #406, opened 2018-07-18
Start of using mwbase instead of pywikibase in wikibase parts

Ladsgroup:
Sitelinks and things related to claims will be complex, let's start with this

Bug: T194758


Ladsgroup:
https://github.com/mediawiki-utilities/python-mwbase/pull/2 will fix this breakage.

Ladsgroup:
It still fails :/

halfak:
Finally, this *should* be ready.

Ladsgroup:
We need to make sure we rebuild our models (and adjust our features in editqaulity and itemquality) before deploying this to prod.

PR #408, opened 2018-08-02
Joblib pickling

adamwight:


adamwight:
This is very interesting,
> joblib also makes it possible to memory map the data buffer of a uncompressed joblib-pickled numpy array when loading it which makes it possible to share memory between processes.

Note that my patch won't do this cos it's currently compressing.

adamwight:
Benchmarks: http://gael-varoquaux.info/programming/new_low-overhead_persistence_in_joblib_for_big_data.html

halfak:
I dug into this and left some notes on the phab task.  Essentially, we shouldn't use joblib, we should instead use cloudpickle, the library that joblib uses.

PR #409, opened 2018-08-07
API changes to use the new MCR-aware params and format

adamwight:
Bug: T201332


halfak:
Looks good to me but I don't have time to test it.

halfak:
Confirmed that it works in practice!

PR #419, opened 2018-12-21
Make basic datasources all json-serializable

Ladsgroup:
Bug: T206333

### DIFF
--- a/revscoring/datasources/meta/timestamp.py
+++ b/revscoring/datasources/meta/timestamp.py
@@ -1,0 +1,30 @@
+"""
+These meta-datasources operate on :class:`revscoring.Datasource`'s that
+return `mwtypes.Timestamp` of the given string.
+
+.. autoclass:: revscoring.datasources.meta.timestamp.Timestamp
+"""
+import mwtypes
+
+from ..datasource import Datasource
+
+MW_REGISTRATION_EPOCH = '2006-01-01T00:00:00Z'
+
+
+class Timestamp(Datasource):
+    """
+    Generates a mwtypes.Timestamp of the given string
+
+    :Parameters:
+        timestamp_str : `str`
+            Timestamp string in ISO format.
+        name : `str`
+            A name for the datasource.
+    """
+
+    def __init__(self, timestamp_str, name=None):
+        super().__init__(name, self.process,
+                         depends_on=[timestamp_str])
+
+    def process(self, timestamp_str):
+        return mwtypes.Timestamp(timestamp_str or MW_REGISTRATION_EPOCH)
--- a/revscoring/datasources/revision_oriented.py
+++ b/revscoring/datasources/revision_oriented.py
@@ -32,9 +32,11 @@ Supporting classes
     :member-order: bysource
 
 """
+import mwtypes
+
 from ..dependencies import DependentSet
+from ..dependencies.util import or_none
 from .datasource import Datasource
-from .meta.timestamp import Timestamp
 
 
 class Revision(DependentSet):
@@ -57,7 +59,9 @@ class Revision(DependentSet):
         "`int` : Revision ID"
         self.timestamp_str = Datasource(name + ".timestamp_str")
         "`str` : Timestamp the revision was saved in ISO format"
-        self.timestamp = Timestamp(self.timestamp_str, name + ".timestamp")
+        self.timestamp = Datasource(
+            name + ".timestamp", or_none(mwtypes.Timestamp),
+            depends_on=[self.timestamp_str])
         ":class:`mwtypes.Timestamp`: Timestamp the revision was saved"
         self.comment = Datasource(name + ".comment")
         "`str` : The comment saved with the revision"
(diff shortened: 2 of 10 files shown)


halfak:
We should probably call these something like "timestamp_str" and then we can implement "timestamp" on top.

The biggest question I'm considering is whether we should allow dependencies to exist in the basic data types.  E.g. we could have something like this:
```
self.timestamp_str = Datasource(name + ".timestamp_str")
"`str` : Timestamp the revision was saved in ISO format"

self.timestamp = Datasource(
    name + ".timestamp", mwtypes.Timestamp, depends_on=[self.timestamp_str])
":class:`mwtypes.Timestamp` : Timestamp the revision was saved"	         
```

This would ensure that the root is always trivially JSON serializable and it would preserve backwards compatibility and the convenience that mwtypes.Timestamp provides.  What do you think?

Ladsgroup:
> We should probably call these something like "timestamp_str" and then we can implement "timestamp" on top.
> 
> The biggest question I'm considering is whether we should allow dependencies to exist in the basic data types. E.g. we could have something like this:
> 
> ```
> self.timestamp_str = Datasource(name + ".timestamp_str")
> "`str` : Timestamp the revision was saved in ISO format"
> 
> self.timestamp = Datasource(
>     name + ".timestamp", mwtypes.Timestamp, depends_on=[self.timestamp_str])
> ":class:`mwtypes.Timestamp` : Timestamp the revision was saved"	         
> ```
> 
> This would ensure that the root is always trivially JSON serializable and it would preserve backwards compatibility and the convenience that mwtypes.Timestamp provides. What do you think?

I tries to implement your suggestion. Take a look.

revscoring/datasources/meta/timestamp.py
```diff
@@ -0,0 +1,30 @@
...
+.. autoclass:: revscoring.datasources.meta.timestamp.Timestamp
+"""
+import mwtypes
+
+from ..datasource import Datasource
+
+MW_REGISTRATION_EPOCH = '2006-01-01T00:00:00Z'
+
+
+class Timestamp(Datasource):
```
halfak: All other meta datasources are lower case and do something a bit more general.  I wonder what is the advantage to this as opposed to:

```
my_datasource = Datasource("my_datasource", process=mwtypes.Timestamp, depends_on=[my_datasource])
```

revscoring/datasources/revision_oriented.py
```diff
@@ -54,8 +55,10 @@ def __init__(self, name,
 
         self.id = Datasource(name + ".id")
         "`int` : Revision ID"
-        self.timestamp = Datasource(name + ".timestamp")
-        ":class:`mwtypes.Timestamp` : Timestamp the revision was saved"
+        self.timestamp_str = Datasource(name + ".timestamp_str")
+        "`str` : Timestamp the revision was saved in ISO format"
+        self.timestamp = Timestamp(self.timestamp_str, name + ".timestamp")
+        ":class:`mwtypes.Timestamp`: Timestamp the revision was saved"
```
halfak: We don't implement any dependencies in datasources.revision_oriented except for this one.  It seems to break the pattern.  I think we can have `timestamp` be here and `timestamp_str` exist inside of the dependencies injected by `api.Extractor`.

revscoring/extractors/api/revision_oriented.py
```diff
@@ -18,9 +16,8 @@ def __init__(self, revision, extractor, rev_doc, id_datasource=None):
 
         self.id = id_datasource or key('revid', rev_doc, name=revision.id.name)
 
-        self.timestamp = key('timestamp', rev_doc,
-                             name=revision.timestamp.name,
-                             apply=mwtypes.Timestamp)
+        self.timestamp_str = key('timestamp', rev_doc,
+                                 name=revision.timestamp.name)
```
halfak: We could implement the timestamp dependency on timestamp_str here.

halfak:
It looks like I was confusing with my last general message about timestamp_str.  I'll take another pass on this change to implement the solution I had in mind.

halfak:
Actually after digging for a while, I think this might be the best option after all.   I'm still looking into good alternatives, but there may be none better than putting "timestamp_str" in `revision_oriented`

revscoring/features/temporal/tests/test_revision_oriented.py
```diff
@@ -4,7 +4,10 @@
 
 from ....datasources import revision_oriented
 from ....dependencies import solve
-from ..revision_oriented import MW_REGISTRATION_EPOCH, revision
+from ..revision_oriented import revision
+
+
+MW_REGISTRATION_EPOCH = Timestamp("2006-01-01T00:00:00Z")
```
halfak: ???

revscoring/features/temporal/revision_oriented.py
```diff
@@ -1,15 +1,12 @@
...
 
-import mwtypes
-
 from pytz import utc
 
 from ...datasources import revision_oriented
 from ...dependencies import DependentSet
 from ..feature import Feature
 
-MW_REGISTRATION_EPOCH = mwtypes.Timestamp("2006-01-01T00:00:00Z")
```
halfak: Why?

halfak:
I pushed a change, but it didn't do very much.  I fixed up some feature naming stuff in the extractor datasources.  I also put the MW_REGISTRATION_EPOCH back the way it was before and moved a nice utility function into dependencies ("or_none").

PR #422, opened 2019-01-23
Allow API extractor to fetch deleted revisions

mariushoch:
Superficially tested


mariushoch:
I need this to run `revscoring extract` for draftquality (which features a lot of deleted revisions in its dataset).

PR #425, opened 2019-01-28
Use sklearn.model_selection instead of sklearn.cross_validation

Ladsgroup:
It's deprecated, see https://github.com/udacity/machine-learning/issues/202

Bug: T192412


Ladsgroup:
BTW: https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html

PR #423, opened 2019-01-23
api extractor: Don't try to "set" & "list"

mariushoch:
TypeError: unsupported operand type(s) for &: 'set' and 'list'

Caused by c1da24de09abc3d68275524caefbbb5666309ad2

PR #427, opened 2019-01-30
Make sure all_dependents variable is being used as a set

Ladsgroup:
Bug: T206333

### DIFF
--- a/revscoring/extractors/api/extractor.py
+++ b/revscoring/extractors/api/extractor.py
@@ -107,7 +107,7 @@ class Extractor(BaseExtractor):
         errored = {}
 
         # Build up caches for data that can be queried in batch
-        if self.revision & all_dependents:
+        if self.revision & set(all_dependents):
             rvprop = set(REV_PROPS)
             if self.revision.text in all_dependents:
                 rvprop.add('content')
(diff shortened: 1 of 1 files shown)


mariushoch:
I have an alternative up at #423 where I switched these from using the bitwise `&` to `and`.

Ladsgroup:
I'm not sure it would work. Generally speaking, and is different from & between sets. E.g. 
```
self.revision = set(123)
all_dependents = set(234)
```
Will give true for "and" and give false for &. I don't know exact values of these variables to judge fully though so I went with conservative approach.

halfak:
I'm confused.  It seem that `all_dependents` is turned into a list at https://github.com/wikimedia/revscoring/blob/e46c3fcbeeba9a348aa3e9b0fb7098e10dc70a23/revscoring/extractors/api/extractor.py#L97

Then in this version of the code, we regularly convert it back to a set in order to perform set operations.  Why not just leave it as a set and convert it to a list when a list is needed?  E.g. https://github.com/wikimedia/revscoring/blob/e46c3fcbeeba9a348aa3e9b0fb7098e10dc70a23/revscoring/extractors/api/extractor.py#L219

Ladsgroup:
Hmm, that sounds okay to me. Made #429

halfak:
@Ladsgroup please delete if you're done with this.

PR #428, opened 2019-02-01
Properly set Datasource/Feature names for b/c aliases

mariushoch:
Otherwise DependentSet will complain about the duplicate names.

https://phabricator.wikimedia.org/T214979


halfak:
This is a good way to get rid of the warnings.  I don't like how it is substantially more verbose.  The features are intended to be identical and that's on purpose.  BUT, I don't see a better option right now.  Maybe we could create a special datasource/feature called `Identity()` that only take a dependent and a new name for the dependent value.  Either way, this is good for now.

PR #430, opened 2019-02-04
Turn members of dependents to string value

Ladsgroup:
Revscoring still understands it

Bug: T206333


halfak:
```
>>> from revscoring.datasources import revision_oriented as ro
>>> ro.revision
{revision}
>>> ro.revision.text in ro.revision
True
>>> str(ro.revision.text) in ro.revision
True
>>> {ro.revision.text} & set(ro.revision)
{<datasource.revision.text>}
>>> {str(ro.revision.text)} & set(ro.revision)
{'datasource.revision.text'}
```

Looks good to me.  The strings should be interchange-able so long as we're not expecting to solve anything in "all_dependents" which I'm pretty sure we're not.

PR #442, opened 2019-07-03
updated the versions of numpy,scipy & scikit-learn

urstrulykkr:
I had updated according to the latest versions of those specified libraries, according to SemVer.
Requesting to get it review my commit.


urstrulykkr:
@halfak  I made my patch, hoping to recieve your kind reviews on it :)

requirements.txt
```diff
@@ -12,12 +12,12 @@ mwtypes >= 0.2.0, < 0.3.999
 mwparserfromhell >= 0.5.1, < 0.5.999
 mysqltsv >= 0.0.7, < 0.0.999
 nltk >= 3.0.0, < 3.0.999
-numpy >= 1.14.0, < 1.14.999
+numpy >= 1.16.2, < 2.0.999
```
halfak: I think we'll want to stay within minor version for these ranges.  E.g. `numpy >= 1.16.2, < 1.6.999`.  See https://semver.org/ for the meaning of these numbers.

urstrulykkr: cool. Will apply the review made :)

urstrulykkr: In here `E.g. numpy >= 1.16.2, < 1.6.999` it should be  <1.16.999 right ?

urstrulykkr:
Thanks a lot for your patience and time.
That makes my 3rd patch to Wikimedia :)

Best
Kaushik.

PR #456, opened 2019-11-21
Add two bad words in languages/serbian.py

zoranzoki21:


halfak:
Looks great!  Thank you.

PR #457, opened 2019-11-21
Update README.md

zoranzoki21:
For installation on ubuntu and debian I added sudo in line 42. There is not difference with line 41, needs root in each case for using apt-get :)


zoranzoki21:
@halfak Can you approve this if is ok?

halfak:
Merged!  Sorry for the delay.

PR #459, opened 2019-11-27
Update english.py

jeblad:
These are strings, and will not parse as regular expressions as they are, but they are later wrapped in regexps so it is probably a matter of preferences and readability.


halfak:
Thank you!

PR #463, opened 2019-12-18
Handles formatting for rates model_info when there are lots of labels.

halfak:
This fixes the formatting of "Rates" classification statistics so that it renders legibly when we have lots of labels.  This is important for our topic models.

### DIFF
--- a/tests/scoring/statistics/classification/tests/test_rates.py
+++ b/tests/scoring/statistics/classification/tests/test_rates.py
@@ -29,3 +29,29 @@ def test_population():
     assert r.lookup("sample.false") == 0.5
     assert r.lookup("population.true") == 0.05
     assert r.lookup("population.false") == 0.95
+
+
+def test_wide_labels():
+    long_counts = Counts(
+        ["I have a bunch of characters",
+         "And I'm going to take up way too much space",
+         "This is definitely going to be too long"],
+        [({'prediction': "I have a bunch of characters"}, "I have a bunch of characters")] * 30 +
+        [({'prediction': "I have a bunch of characters"}, "And I'm going to take up way too much space")] * 2 +
+        [({'prediction': "I have a bunch of characters"}, "This is definitely going to be too long")] * 1 +
+        [({'prediction': "And I'm going to take up way too much space"}, "And I'm going to take up way too much space")] * 25 +
+        [({'prediction': "And I'm going to take up way too much space"}, "I have a bunch of characters")] * 5 +
+        [({'prediction': "And I'm going to take up way too much space"}, "This is definitely going to be too long")] * 6 +
+        [({'prediction': "This is definitely going to be too long"}, "This is definitely going to be too long")] * 35 +
+        [({'prediction': "This is definitely going to be too long"}, "I have a bunch of characters")] * 1 +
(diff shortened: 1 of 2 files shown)


halfak:
When we have few labels, the output looks like this:
```
rates:
	              True    False
	----------  ------  -------
	sample        0.5      0.5
	population    0.05     0.95
```

When we have wide/many labels, the output looks like this: 

```
rates:
	                                               sample    population
	-------------------------------------------  --------  ------------
	I have a bunch of characters                    0.34          0.45
	And I'm going to take up way too much space     0.264         0.332
	This is definitely going to be too long         0.396         0.25
```

tests/scoring/statistics/classification/tests/test_rates.py
```diff
@@ -29,3 +29,29 @@ def test_population():
     assert r.lookup("sample.false") == 0.5
     assert r.lookup("population.true") == 0.05
     assert r.lookup("population.false") == 0.95
+
+
+def test_wide_labels():
```
kevinbazira: At the bottom of `test_rates.py`, I added;

```
# test_simple()
# test_population()
# test_wide_labels()
```


and run;

```
python test_rates.py 
```


with each of the lines above uncommented in intervals, both `test_simple()` and `test_population()` returned the expected result but `test_wide_labels()` returned;

```
rates:
	              'I have a bunch of characters'    "And I'm going to take up way too much space"    'This is definitely going to be too long'
	----------  --------------------------------  -----------------------------------------------  -------------------------------------------
	sample                                  0.34                                            0.264                                        0.396
	population                              0.45                                            0.332                                        0.25

Traceback (most recent call last):
  File "test_rates.py", line 55, in <module>
    test_wide_labels()
  File "test_rates.py", line 51, in test_wide_labels
    assert len(line) < 80
AssertionError
```


was the output above the expected result or the one below is the expected result? (based on this comment: https://github.com/wikimedia/revscoring/pull/463#issuecomment-567109024)

```
rates:
	                                               sample    population
	-------------------------------------------  --------  ------------
	I have a bunch of characters                    0.34          0.45
	And I'm going to take up way too much space     0.264         0.332
	This is definitely going to be too long         0.396         0.25
```

halfak: I'm not sure what you're seeing in your test, but it looks like travis CI agrees that the test passes.  Maybe you're inadvertently executing this test code with an old version of revscoring.

halfak: To be clear, I recommend running tests using "pytest" from the base of the repo.  You can run this test specifically by running:

```
$ pytest tests/scoring/statistics/classification/tests/test_rates.py
```

kevinbazira: > I'm not sure what you're seeing in your test, but it looks like travis CI agrees that the test passes. Maybe you're inadvertently executing this test code with an old version of revscoring.

Yep, I was running an old version of revscoring.


> To be clear, I recommend running tests using "pytest" from the base of the repo. You can run this test specifically by running:
> 
> ```
> $ pytest tests/scoring/statistics/classification/tests/test_rates.py
> ```

Thanks for the recommendation. I run it with `pytest` and all the 3 tests passed.


Otherwise, this LGTM and I'm going to merge it now.

PR #465, opened 2019-12-23
Generate word vectors using fasttext

kevinbazira:
This has been based on the vector models on stat1007:

Forexample, go to;

```
$ cd /home/kevinbazira/fasttext_word_representations/2__unsupervised-tutorial
```

Then run Python;

```
$ Python
```

And add;

```
>>>import fasttext

>>>def vectorize_words(vector_model, words):
    	model = fasttext.load_model(vector_model)
    	return [model.get_word_vector(x) for x in words]

>>>vectorize_words("enwiki_latest_pages_articles_result_skipgram_100.bin", ["asparagus", "pidgey", "yellow"])
...
```

This is the `vectorize_words` method I've added to the `FastText` class in `vectorizers.py`


revscoring/datasources/meta/vectorizers.py
```diff
@@ -62,3 +64,29 @@ def load_kv(filename=None, path=None, limit=None):
...
+            A name for the `revscoring.FeatureVector`
+    """  # noqa
+
+    def __init__(self, words_datasource, vectorize_words, name=None):
+        name = self._format_name(name, [words_datasource, vectorize_words])
+
+        super().__init__(name, vectorize_words, depends_on=[words_datasource])
+
+    def vectorize_words(vector_model, words):
+        model = fasttext.load_model(vector_model)
```
halfak: This is a big operation.  It'll run every time we want to process a set of words.  I it takes 3-4 seconds on my computer.

kevinbazira: Yep it's a big operation indeed. As you advised earlier, I've worked on a new update that re-uses the word2vec datasource since the fasttext `.vec` files are "word2vec format".

revscoring/datasources/meta/vectorizers.py
```diff
@@ -6,42 +6,41 @@
...
             a function that takes a list of words and converts it to a list
             of vectors of those words
+        keyed_vectors : `dict`
+            A list of word vectors in "word2vec format"
         name : `str`
             A name for the `revscoring.FeatureVector`
     """  # noqa
 
-    def __init__(self, words_datasource, vectorize_words, name=None):
+    def __init__(self, words_datasource, vectorize_words, keyed_vectors, name=None):
```
halfak: keyed_vectors should be stored in the object.  E.g. `self.keyed_vectors == keyed_vectors`

halfak: We should also drop the "vectorize_words" argument.  We'll implement it natively.

revscoring/datasources/meta/vectorizers.py
```diff
@@ -6,42 +6,41 @@
...
     """  # noqa
 
-    def __init__(self, words_datasource, vectorize_words, name=None):
+    def __init__(self, words_datasource, vectorize_words, keyed_vectors, name=None): 
         name = self._format_name(name, [words_datasource, vectorize_words])
 
         super().__init__(name, vectorize_words, depends_on=[words_datasource])
 
     @staticmethod
     def vectorize_words(keyed_vectors, words):
```
halfak: Let's change this to a regular instance method. 

```
def process(self, words):
    if not words:
        return [[0] * self.keyed_vectors.vector_size]
    ...
```

This will mean that we change the usage to look like this:
```
keyed_vectors = {...}
w2v = word2vec(words_datasource, keyed_vectors)
solve(w2v, cache={words_datasource: ["I", "am", "some", "words"])
```

PR #466, opened 2020-01-03
English idioms

HAKSOAT:
Add revscoring.languages.english.get_idioms method

Idioms can be problematic in formal articles. So,
a list of idioms are needed to track them when they appear
in the articles.

Task: [T205545](https://phabricator.wikimedia.org/T205545)


revscoring/languages/english.py
```diff
@@ -264,3 +266,23 @@ def safe_dictionary_check(word):
...
+        list='categorymembers',
+        cmtitle="Category:English idioms",
+        formatversion=2,
+        continuation=True)
+
+    idioms = []
+    for doc in results:
+        for page_doc in doc['query']['categorymembers']:
+            idiom = page_doc['title']
+            if 'Category:English' not in idiom:
```
halfak: What's the meaning of this check?

HAKSOAT: In the last five or so results.... There are values which are not idioms that have Category:English as suffixes. That's why I checked to ensure they are not added to the list.

halfak:
This is great.  Thanks for your work.  I think we'll want to turn this into an asset that we can load in the enwiki.py file rather than making an API request every time we load the file.  

Maybe you could make a utility in `revscoring/utilities` that prints out the list of idioms.  Then we can turn that into a file we load in enwiki.py.  What do you think?

HAKSOAT:
I agree. I wasn't comfortable putting that function there in the first place.
 Just had to do something and get feedback. I'll look into your suggestion and get back to you.

revscoring/languages/english.py
halfak: Aha!  That makes sense.  It might me work adding a comment explaining this.  Otherwise, it looks good.

halfak:
Sounds great!  Thanks for your work :)

HAKSOAT:
Hello @halfak How about the idioms are extracted and redirected into a text file, such that we have an idiom per line and can read that data into memory when needed.

There are 7943 idioms in the list.

halfak:
That sounds great.

HAKSOAT:
I think I just messed up the PR. I can see commits I didn't make.

HAKSOAT:
Hello @halfak The idioms are in the revscoring/utilities folder. Stored as idioms.txt.

halfak:
Yeah.  It looks like you might need a rebase.  I can dive into this to get it cleaned up, but I'm not sure I'll have time for that soon.  Alternatively, you could start a fresh PR and copy-paste your work.

HAKSOAT:
Hello @halfak I have resolved the issue.

halfak:
Nice work!  This looks great.  Could you try to adapt your script in "utilities" to look like the others.  Once you do, you can run your script from the base directory by running:
```
./utility fetch_idioms
```

You'll note that the other utilities write to sys.stdout by default.  I'd suggest copy-pasting that boilerplate.  Note that we're using `docopt` in the other utilities.  Docopt parses the script documentation at the top of the page in order to figure out how to process the arguments.

HAKSOAT:
Okay. I'll work on that. Thanks.

revscoring/utilities/fetch_idioms.py
```diff
@@ -0,0 +1,79 @@
+"""
+``revscoring fetch_idioms -h``
+::
+
+    Gets a list of idioms.
+
+    Usage:
+        fetch_text [--output=<path>]
```
halfak: fetch_idioms

revscoring/utilities/fetch_idioms.py
```diff
@@ -0,0 +1,79 @@
+"""
+``revscoring fetch_idioms -h``
+::
+
+    Gets a list of idioms.
```
halfak: Gets a list of *English language* idioms from en.wiktionary.org

revscoring/utilities/fetch_idioms.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Usage:
+        fetch_text [--output=<path>]
+                   [--verbose] [--debug]
+
+    Options:
+        -h --help        Print this documentation
+        --output=<path>  Path to a file to write the idioms
+                         [default: revscoring/utilities/idioms.txt]
```
halfak: Should default to <stdout> like the other utilities.

revscoring/utilities/fetch_idioms.py
```diff
@@ -0,0 +1,79 @@
...
+
+def save(file_path, idioms):
+    with open(file_path, 'w') as file:
+        for idiom in idioms:
+            file.write(idiom + "\n")
+
+
+def run(output, verbose):
+    if verbose:
+        sys.stdout.write('Fetching idioms...')
```
halfak: You probably want something like:
```
logger.info("Fetching idioms...")
```

revscoring/utilities/fetch_idioms.py
```diff
@@ -0,0 +1,79 @@
...
+    idioms = fetch()
+    if verbose:
+        sys.stdout.write('Saving idioms...')
```
halfak: Same here. 
```
logger.info("Saving idioms...") 
```

halfak:
Looks good.  I left a few notes.  I think we'll want to move the output to a new directory.  Maybe we can put it into the base directory and call it "asserts/".

HAKSOAT:
@halfak You mean "assets" instead of "asserts" right?

HAKSOAT:
I've made changes. I used "assets" instead of "asserts".

halfak:
Yes!  Thank you.

halfak:
Some notes from IRC: 

We need to first add the idioms to revscoring/languages/english.py

Check out what we do with badwords and informals for an example for how we can construct a set of "RegexMatches" features. 

You should be able to load in the idioms, construct the RegexMatches object and then we can use that to scan for idioms are use those counts as features. 

Once we get all of that worked out, we'll turn to the model repos like editquality, articlequality, etc.

halfak:
I merged this manually.  not sure why github didn't notice.

PR #472, opened 2020-02-14
Feature extraction images

HAKSOAT:
Add revscoring.features.wikitext.datasources.Revision.templates_with_name_matching method
Add revscoring.features.wikitext.datasources.Revision.tags_with_name_matching method

Images are only counted when they are in the format: [[File:Something.jpg]].
However, there are other ways of adding images into tags and templates. The tags and templates of concern need to be extracted, so that further parsing can be done.

Task: T180822

### DIFF
--- a/revscoring/features/wikitext/datasources/parsed.py
+++ b/revscoring/features/wikitext/datasources/parsed.py
@@ -105,6 +105,14 @@ class Revision:
         Returns a list of html tag names present in the content of the revision
         """
 
+        self.tags_str = mappers.map(
+            str, self.tags,
+            name=self._name + ".tags_str"
+        )
+        """
+        Returns a list of tags present in the content of the revision as strings
+        """
+
         self.templates = get_key(
             mwparserfromhell.nodes.Template, self.node_class_map,
             default=[],
@@ -120,6 +128,14 @@ class Revision:
         )
         """
         Returns a list of template names present in the content of the revision
+        """
+
+        self.templates_str = mappers.map(
+            str, self.templates,
+            name=self._name + ".templates_str"
+        )
+        """
+        Returns a list of templates present in the content of the revision as strings
         """
 
     def heading_titles_matching(self, regex, name=None):
@@ -190,6 +206,20 @@ class Revision:
 
         return filters.regex_matching(regex, self.tag_names, name=name)
 
+    def tags_str_matching(self, regex, name=None):
+        """
+        Constructs a :class:`revscoring.Datasource` that returns all tags
+        that matches a regular expression as strings.
+        """
+        if not hasattr(regex, "pattern"):
+            regex = re.compile(regex, re.I)
+
+        if name is None:
+            name = "{0}({1})" \
+                   .format(self._name + ".tags_str_matching", regex.pattern)
+
+        return filters.regex_matching(regex, self.tags_str, name=name)
+
     def template_names_matching(self, regex, name=None):
         """
         Constructs a :class:`revscoring.Datasource` that returns all template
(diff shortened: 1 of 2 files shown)


halfak:
I like this.  It's a powerful way to extract information about what is contained *inside of* tags/templates.  

I think we can do this a bit more simply and I don't like putting "html" into the name because it isn't totally accurate since we're looking at wiki markup.  For the templates, there's likely no html present!

I think we can get the string representation of tags/templates by doing something like this:

```
from revscoring.datasources.meta import mappers, filters
from revscoring.features import wikitext

revision_tag_strs = mappers.map(str, wikitext.revision.datasources.tags)
revision_template_strs = mappers.map(str, wikitext.revision.datasources.templates)

my_tags = filters.regex_matching(revision_tag_strs, r"some cool regex")
my_templates = filters.regex_matching(revision_template_strs, r"some cool regex") 
```

What do you think?

halfak:
Oh and just so you know the build was failing because of a style issue.  Here's the interesting lines:

```
$ flake8 . --max-line-length=85 --exclude=.svn,CVS,.bzr,.hg,.git,__pycache__,.tox,.eggs,*.egg,docs
./tests/features/wikitext/tests/test_parsed.py:164:74: W292 no newline at end of file
The command "flake8 . --max-line-length=85 --exclude=.svn,CVS,.bzr,.hg,.git,__pycache__,.tox,.eggs,*.egg,docs" exited with 1.
```

HAKSOAT:
> 
> 
> I like this. It's a powerful way to extract information about what is contained _inside of_ tags/templates.
> 
> I think we can do this a bit more simply and I don't like putting "html" into the name because it isn't totally accurate since we're looking at wiki markup. For the templates, there's likely no html present!
> 
> I think we can get the string representation of tags/templates by doing something like this:
> 
> ```
> from revscoring.datasources.meta import mappers, filters
> from revscoring.features import wikitext
> 
> revision_tag_strs = mappers.map(str, wikitext.revision.datasources.tags)
> revision_template_strs = mappers.map(str, wikitext.revision.datasources.templates)
> 
> my_tags = filters.regex_matching(revision_tag_strs, r"some cool regex")
> my_templates = filters.regex_matching(revision_template_strs, r"some cool regex") 
> ```
> 
> What do you think?

This is beautiful. I never thought about this. Takes away the need for the extract functions.

HAKSOAT:
I'll make the required changes and push. Also, do you think the parsing of the tag and template should be done in revscoring? I'm thinking of doing it in the article quality repo.

revscoring/features/wikitext/datasources/parsed.py
```diff
@@ -190,6 +206,20 @@ def tag_names_matching(self, regex, name=None):
 
         return filters.regex_matching(regex, self.tag_names, name=name)
 
+    def tags_str_with_name_matching(self, regex, name=None):
+        """
+        Constructs a :class:`revscoring.Datasource` that returns all tags having a name
+        that matches a regular expression as strings.
+        """
+        if not hasattr(regex, "pattern"):
+            regex = "<{}.*>".format(regex)
```
halfak: I'm not sure I like this strategy.  I think we should either allow the regex match the complete tag and content or we should use mwparserfromhell to extract the tag name to match against.

halfak:
+1  I think actually matching stuff belongs in the articlequality repo.  I just left one note.  Otherwise this looks good to me.

halfak:
Oh!  There should also be tests for the regex matching methods.

HAKSOAT:
Hello @halfak I have made changes. Next will be to work on the articlequality repo. But the repo seems to be dependent on revscoring from the pypi. Does it get updated after merging a PR?

halfak:
Nice work!  We'll need to make a new release of revscoring and then you'll want to update the requirements.txt there.  I'll look into making a release today.

halfak:
I just released revscoring 2.6.6.  You should be able to use that to take the next step in the articlequality repo.

HAKSOAT:
Great. I'll proceed then. Thanks.

HAKSOAT:
Hello @halfak I am trying to connect to the IRC channel, only to find out that I am banned. I have no idea what I did wrong.

PR #474, opened 2020-02-29
Added _extract_sections() method in features/wikitext/datsources/pars…

chtnnh:
…ed.py and self.sections which returns sections

This enables dividing an article/revision into sections, which we require to judge text complexity as features in article quality. 

@halfak

### DIFF
--- a/revscoring/features/wikitext/datasources/parsed.py
+++ b/revscoring/features/wikitext/datasources/parsed.py
@@ -137,6 +137,13 @@ class Revision:
         """
         Returns a list of templates present in the content of the revision as strings
         """
+        self.sections = execute_method(
+            "_extract_sections", self.wikicode,
+            name=self._name + ".content"
+        )
+        """
+        Returns list of sections in the article as wikicode shared node list
+        """
 
     def heading_titles_matching(self, regex, name=None):
         """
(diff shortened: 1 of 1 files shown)


revscoring/features/wikitext/datasources/parsed.py
```diff
@@ -286,6 +293,9 @@ def _extract_tag_name(tag):
 def _extract_template_name(template):
     return str(template.name)
 
+def _extract_sections(template):
+    return mwparserfromhell.wikicode.Wikicode.get_sections(flat=True).template
```
halfak: Why does this line end in ".template"?

chtnnh: We are calling get_sections on template.

halfak: Looks to me like you are doing something a bit too complicated.  I think you mean to do something like: 
```
        self.sections = Datasource(
            self._name + ".sections",
            _extract_sections, depends_on=[self.wikicode]
        )
```

and 

```
def _extract_sections(wikicode):
    return wikicode.filter_sections(flat=True)
```

revscoring/features/wikitext/datasources/parsed.py
```diff
@@ -286,6 +293,9 @@ def _extract_tag_name(tag):
 def _extract_template_name(template):
     return str(template.name)
```
halfak: Flake8 will complain that there needs to be two empty lines before top-level function definitions.  Otherwise, this looks good to me.

chtnnh: Oh, I totally forgot about that
Let me edit and push again?

PR #475, opened 2020-03-05
Added class not_none to revscoring/datasources/meta

chtnnh:
to filter None items in list.

@halfak As discussed, here is the PR.

### DIFF
--- a/revscoring/datasources/meta/filters.py
+++ b/revscoring/datasources/meta/filters.py
@@ -114,7 +114,7 @@ class not_none(filter):
     """
 
     def __init__(self, items_datasource, name=None):
-        name = self.format_name(name, [items_datasource])
+        name = self._format_name(name, [items_datasource])
         super().__init__(self.is_not_none, items_datasource, name=name)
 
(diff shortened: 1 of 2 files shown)


revscoring/datasources/meta/filters.py
```diff
@@ -100,3 +100,23 @@ def __init__(self, numbers_datasource, name=None):
...
+            A name for the datasource.
+    """
+
+    def __init__(self, items_datasource, name=None):
+        name = self.format_name(name, [items_datasource])
+        super().__init__(self.is_not_none, items_datasource, name=name)
+
+    def is_not_none(self, v):
+        if v is not None:
+            return v
```
halfak: This should return a boolean.  I think you just want to return `v is not None`

chtnnh: Wait, it returns v if v is not None

if we return v is not None, that will be the boolean right?
Please confirm

halfak:
LGTM. Just waiting on travis.

PR #477, opened 2020-03-09
Filter List of Idioms

HAKSOAT:
Modify revscoring.utilities.fetch_idioms.fetch method
Modify revscoring.assets.enwiktionary_idioms file

The idioms gotten from Enwikitionary Category:English idioms,
have problematic idioms, with many of them being regular
English phrases instead of idioms. The list was also quite
long, causing it to take a long time when matching idioms.
Idioms can be problematic in formal articles.
Hence, as many regular phrases as possible have to be removed, and
the list should be reduced too.

Task: T247000


revscoring/utilities/fetch_idioms.py
```diff
@@ -55,15 +56,49 @@ def fetch(output):
...
+            first_term = terms[0]
+            other_terms = [" ".join(terms[1:])]
+
+    if len(other_terms) < 2:
+        idiom_regex = '{} {}'.format(
+            first_term, "".join(other_terms))
+    else:
+        idiom_regex = '{} ({})'.format(
+            first_term, "|".join(other_terms))
+    if len(idiom_regex):
```
halfak: How could this ever return false?  I think that `bool(len(" "))` is True.  That's the most *falsey* thing I could construct with `idiom_regex = '{} {}'.format(first_term, "".join(other_terms))`

revscoring/utilities/fetch_idioms.py
```diff
@@ -55,15 +56,49 @@ def fetch(output):
...
+        # Remove one word idioms
+        if re.match(r'^[\w-]+$', idiom):
+            continue
+        # Remove two-worded idioms
+        if re.match(r'^[\w-]+ [\w-]+$', idiom):
+            continue
+        # Remove similes
+        if 'as a' in idiom:
+            continue
+        terms = idiom.split()
```
halfak: It seems like this section should be another function.  like:

```
def regexify(idioms):
  ...
```

It would return a generator that produces the regexes you'd like to output.  This will be handy because it'll make testing a lot easier.

revscoring/utilities/fetch_idioms.py
HAKSOAT: I missed that ... it used to be bool(len("")), but I added a space later on. I'll make changes.

halfak:
Looks great!  Nice work

PR #483, opened 2020-04-08
Add words_to_watch for pt

he7d3r:
Adapted from [w:pt:Wikipédia:Palavras a evitar](https://pt.wikipedia.org/wiki/Wikipedia:Palavras_a_evitar?oldid=55314927).


halfak:
We'll need some tests to make sure we can get these phrases out of the text.  See https://github.com/wikimedia/revscoring/blob/master/tests/languages/test_english.py#L238 for an example of doing it for English.

PR #485, opened 2020-04-17
Fix broken fetch_text when access deleted pages.

halfak:


halfak:
This fix is pretty minor and obvious so I'm self-merging.

PR #488, opened 2020-05-16
Improve English Idioms

HAKSOAT:
Add revscoring.utilities.fetch_idioms.convert_trie_to_regex method
Add revscoring.utilities.fetch_idioms.construct_trie method
Add revscoring.utilities.fetch_idioms.create_regex method
Modify revscoring.assets.enwiktionary_idioms.txt

    The list of idioms being used has a lot of regular phrases that
    are not idioms. It also takes a long time to run when finding
    matches in large text.

    The regex generated from the list of idioms simply pipes each idiom
    together. It can become much more efficient by factoring out common
    characters and using pipes on the non-common characters or group of
    characters.

Task: T247000


HAKSOAT:
HEADS-UP: @halfak I had to change Python3.5 on Travis to 3.6. This is because I used dictionaries for my trie implementation. In 3.5 dictionaries are not orderly, hence it is causing tests to fail.

revscoring/utilities/fetch_idioms.py
```diff
@@ -63,48 +64,46 @@ def is_idiom(phrase):
...
-    if not idiom_regex.isspace():
-        return idiom_regex
+        combo = []
+        # Traverse depth first
+        for key, value in zip(keys, values):
+            combo.append(key + convert_trie_to_regex(value))
+        if len(combo) > 1:
+            return "(?:{})".format('|'.join(combo))
+        else:
+            return ''.join(combo)
```
he7d3r, line 97: Shouldn't we apply `re.escape(...)` somwhere?

revscoring/assets/enwiktionary_idioms.txt
```diff
@@ -381,131 +1147,696 @@
...
-"four (score and seven years ago|sheets to the wind)"
+"four score and seven years ago"
+"four sheets to the wind"
 "fox in the henhouse"
-"fresh (off the boat|out of)"
+"fresh off the boat"
+"fresh out of"
 "fret the gizzard"
-"friend (of Bill W.|with benefits)"
+"friend of Bill W."
```
he7d3r, line 1224: Will this `.` be escaped afterwards?

HAKSOAT: Hello @he7d3r Thanks for pointing out. I'll separate those.

revscoring/utilities/fetch_idioms.py
HAKSOAT: @he7d3r My first approach to this was to generate a trie out of the regular list, then convert it into optimized regex instead of piping the words together. Hence, with that approach, we can't allow any regex in the list of words.

revscoring/datasources/meta/extractors.py
```diff
@@ -64,3 +66,41 @@ def process(self, text_or_texts):
...
+    def process(self, text_or_texts):
+        if text_or_texts is None:
+            return []
+        elif isinstance(text_or_texts, str):
+            text = text_or_texts
+            return self.word_processor.extract_keywords(text)
+        else:
+            texts = text_or_texts
+            return [self.word_processor.extract_keywords(text)
+                    for text in texts]
```
halfak, line 106:
```
return [substring
        for text in texts
        for substring in self.word_processor.extract_keywords(text)]
```

HAKSOAT: Hmmmm. How did I miss that? Thank you.

revscoring/utilities/fetch_idioms.py
halfak: It doesn't seem like we need this code anymore, do we?

HAKSOAT: We don't.

PR #493, opened 2020-08-03
corrects the swapped column names

dibyaaaaax:
@halfak while building models in drafttopic, I noticed that column names FP and FN in L80 were swapped. I think this will fix the order.


halfak:
Can you show what was wrong and is now right with this behavior via the test?

dibyaaaaax:
> Can you show what was wrong and is now right with this behavior via the test?

@halfak I have added a test. Not sure if you meant something like that.

halfak:
This looks good.  The tests are failing for some version issue with pytest.  I'm looking into a good solution.

PR #496, opened 2020-08-24
Add page.entity

Ladsgroup:
Similar to page.suggested.properties

It's still not yet ready as wbgetentities doesn't support revids

Bug: T260774


mariushoch:
I guess we want to abandon this now?

Ladsgroup:
> I guess we want to abandon this now?

Yup

PR #497, opened 2020-09-14
Replace references to APIExtractor in docs

paulkernfeld:
The code was previously refactored but a few references in docs were not
updated.
It looks like the "notes on adhoc job" is old and could still be out of
data

### DIFF
--- a/docs/notes_on_adhoc_jobs.txt
+++ b/docs/notes_on_adhoc_jobs.txt
@@ -9,22 +9,18 @@ makes predictions
 Imagine some python (requires `pip install mwreverts`):
 
 # There exists a model file at enwiki.model
-from itertools import groupby
-
 from mwapi import Session
 import mwreverts
 
 from revscoring.extractors.api.extractor import Extractor
 from revscoring.scoring.models import LinearSVC
 
-model = LinearSVC.MODEL.from_file(open("enwiki.model", 'rb'))
-
 session = Session("https://en.wikipedia.org/w/api.php")
 extractor = Extractor(session)
 
-scorer = LinearSVC(extractor, model)
+model = Model.load(open("enwiki.model", "rb"))
 
-api_result = session.get(action='query', titles = 'Main Page', prop='revisions', rvlimit=500, rvprop='sha1')
+api_result = session.get(action='query', titles='Main Page', prop='revisions', rvlimit=500, rvprop='sha1|ids')
 revisions = next(iter(api_result['query']['pages'].values()))['revisions']
 
 # Content that has been revision-deleted has a hidden SHA-1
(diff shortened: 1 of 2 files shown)


revscoring/extractors/extractor.py
```diff
@@ -43,7 +43,7 @@ class OfflineExtractor(Extractor):
     def __init__(self):
         super().__init__()
         logger.warning("Loading OfflineExtractor.  You probably want an " +
-                       "APIExtractor unless this is the test server.")
+                       "extractors.api.extractor.Extractor unless this is the test server.")
```
halfak, line 46: It'd be better to call this `revscoring.extractors.api.Extractor`

docs/notes_on_adhoc_jobs.txt
```diff
@@ -14,13 +14,13 @@ from itertools import groupby
 from mw import api
```
halfak, line 14: No longer using `mw`.  Now we get `Session` from `mwapi`.

docs/notes_on_adhoc_jobs.txt
```diff
@@ -14,13 +14,13 @@ from itertools import groupby
...
 from mw.lib import reverts
 
-from revscores import APIExtractor
-from revscores.scorers import LinearSVC
+from revscoring.extractors.api.extractor import Extractor
+from revscoring.scorers import LinearSVC
```
halfak, line 18: This should be `from revscoring.scoring.models import LinearSVC"

docs/notes_on_adhoc_jobs.txt
```diff
@@ -14,13 +14,13 @@ from itertools import groupby
...
 
 model = LinearSVC.MODEL.from_file(open("ptwiki.model", 'rb'))
 
 session = api.Session("https://pt.wikipedia.org/w/api.php")
```
halfak, line 22:
```
session = mwapi.Session(...)
```

halfak:
Thanks for catching this.  This example had a lot of things wrong with it.  I left a few notes.  Up to you if you want to take them all on.

paulkernfeld:
Thanks for the quick review! I attempted to fix each of the comments above, but I haven't been able to get the project running yet.

halfak:
Are you able to import `revscoring` or are you stuck on an installation/dependency step?  The details in the README should help with the build requirements.

paulkernfeld:
It took me a while to figure out how to get scipy 1.2 going on MacOS Catalina, but I now have that working. I also found that `brew install aspell --with-all-languages` didn't work on my machine so I just did `brew install aspell` 🤷 . 

I see that we're no longer depending on `mediawiki-utilities`, so I'm not sure where to get the functionality to replace `from mw.lib import reverts` in `notes_on_adhoc_jobs.txt`. Should I add a test dependency on `mwreverts`, maybe?

paulkernfeld:
I think I can fix up most of the stuff from `notes_on_adhoc_jobs`, but it would also be helpful to have a sample `.model` file. Is there somewhere I can download a pre-built model? Even a toy model could be useful for smoke-testing the sample script.

paulkernfeld:
I think the notes should be good to go now.

I constructed a dummy model like this for testing:

```python
from revscoring import Feature
from tests.scoring.models.tests.util import process_float
model = LinearSVC([Feature("some_float", process_float(), depends_on=[], returns=float)], [True, False])
model.train([((0,), False), ((1,), True)])
```

accraze (approved):
LGTM :+1: thanks @paulkernfeld

Issue #310, opened 2017-05-03
Install revscoring on OS X


bkowshik:
### 1. Per documentation on README

```sh
$ brew install aspell --with-all-languages
$ brew install enchant

$ mkvirtualenv wikiai_revscoring
pip install --no-binary pyenchant revscoring
...
Collecting scipy<0.17.999,>=0.13.3 (from revscoring)
  Using cached scipy-0.17.1.tar.gz
    Complete output from command python setup.py egg_info:
    Download error on https://pypi.python.org/simple/numpy/: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:749) -- Some packages may not be found!
    Couldn't find index page for 'numpy' (maybe misspelled?)
    Download error on https://pypi.python.org/simple/: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:749) -- Some packages may not be found!
    No local packages or working download links found for numpy>=1.6.2
...
```

### 2. This looks like a known issue!

- [pip install pandas fails to download numpy dependency](https://github.com/pandas-dev/pandas/issues/4452)

Per comment [here](https://github.com/pypa/pip/issues/3238#issuecomment-155345708), installing `numpy` first and then installing `revscoring` is the solution.

```sh
$ pip install numpy

$ pip install --no-binary pyenchant revscoring
Running setup.py bdist_wheel for scikit-learn ... error
Complete output from command /Users/bkowshik/.virtualenvs/wikiai_revscoring/bin/python3.6 -u -c "import setuptools, tokenize;__file__='/private/var/folders/_0/75g04h290zz51r_l4k_f22zr0000gn/T/pip-build-r55j2ov3/scikit-learn/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /var/folders/_0/75g04h290zz51r_l4k_f22zr0000gn/T/tmpdj729dstpip-wheel- --python-tag cp36:
Partial import of sklearn during the build process.
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/private/var/folders/_0/75g04h290zz51r_l4k_f22zr0000gn/T/pip-build-r55j2ov3/scikit-learn/setup.py", line 247, in <module>
    setup_package()
[...]

PR #500, opened 2020-10-12
Delete the model ignore, creating issues with testing.

WhenDayBreaks:
Fixing #310, and editquality does this as well. [ci skip]


halfak:
Removing the .model from the ignore file is not necessary to include models with the repo.  It just requires a conscious action to add the models back.  E.g. `git add models/some_model.model -f`

PR #509, opened 2021-11-04
Score utility now handles rev_docs

halfak:


accraze (commented):
This looks good at a high level, but I'm not sure what this solution is intended for. Can you please provide more info/context/phab task?

halfak:
All of the other utilities read and write rev_docs.  This utility was a hold over from an old way of processing data files.  So this is continuing standardization across our utilities.  Right now, we want to use it to generate lists of false-positives and false-negatives.  Doing so is much easier when we can include labeled data with the observation as well as the score in the output -- which is something the rev_doc data format makes easy through extensible json-lines.

accraze (approved):
:eyes: :+1: 

thanks for adding context!

PR #515, opened 2022-02-08
Bumps requirement for mwparserfromhell to 0.6.x

halfak:


elukey:
LGTM, I'd like Andy and Kevin to check as well just to be sure. IIRC we need 0.6+ to fix the segfault that affected celery, so the range looks good.

kevinbazira (approved):
LGTM too!

PR #517, opened 2022-03-10
Improve error message when decoding a non-item revision from Wikidata

elukey:
Revscoring returns a very verbose and cryptic JSONDecodeError stacktrace
when it tries to score a non-item Wikidata revision with the item
quality model (for example, a User/Talk page change, etc..).
This change adds a new exception meant to capture this kind of error,
to wrap it in a clearer error message.

Co-authored-by: halfak <aaron.halfaker@gmail.com>
Bug: T302851


revscoring/errors.py
```diff
@@ -75,6 +75,11 @@ def __init__(self, datasources, info=None, arg=None):
         super().__init__("Query failed ({0}:{1})"
                          .format(datasources, info))
 
+class UnexpectedContentType(DependencyError):
+    def __init__(self, text, arg=None):
+        super().__init__("Expected entity JSON "
+                         "but the following content can't be parsed: {}"
+                         .format(text))
```
halfak, line 82: I think it would be best not to pre-fill the text of the error to be about JSON given that this is named as a general error for unexpected content.  Or you could make an "EntityContentTypeError" class that extends "UnexpectedContentType" and pre-fills this error message.

revscoring/features/wikibase/datasources/revision_oriented.py
```diff
@@ -114,7 +115,10 @@ def __init__(self, name, revision_datasources):
 
 def _process_entity_doc(text):
     if text is not None:
-        return json.loads(text)
+        try:
+            return json.loads(text)
+        except json.decoder.JSONDecodeError:
+            raise UnexpectedContentType(text)
```
halfak, line 121: To be clear, I recommend saying something specific about "Entity JSON" here.

OR raising an error that is more specific to Entity JSON parsing that extends UnexpectedContentType

Given that I think this is the only place we might throw such an error, I'm more in favor of the former, but I could see value in the latter.  It's nice to have the consistent parts of error text defined in errors.py

revscoring/errors.py
halfak: Oh!  One more note.  `text` can be very long so it would be good to trim it down to the first N chars.  20 or 50 chars should be plenty for debugging.

elukey: Good point, lemme know if the new version is better!

revscoring/features/wikibase/datasources/revision_oriented.py
elukey: Definitely, I added a specific parameter to the Exception, lemme know what you think about it.

halfak (approved):
Looks great.  Nice work.

revscoring/errors.py
```diff
@@ -75,6 +75,12 @@ def __init__(self, datasources, info=None, arg=None):
         super().__init__("Query failed ({0}:{1})"
                          .format(datasources, info))
 
+class UnexpectedContentType(DependencyError):
+    def __init__(self, text, content_type, arg=None):
+        super().__init__("Expected entity of content type {} "
```
halfak, line 80: I approved this but I just noticed that "Entity" in this message.  Entity is jargon for a specific type of wiki content.  Entity JSON is the format that Items, Properties, and Lexemes are built on top of for Wikidata.  Maybe this could be "Expected content of type {}" where the value could be "Entity JSON".

halfak (changes requested):
Just caught something that might be confusing down the road.

revscoring/errors.py
elukey: Good point, amended, let me know if it is better now :)

revscoring/errors.py
```diff
@@ -75,6 +75,12 @@ def __init__(self, datasources, info=None, arg=None):
         super().__init__("Query failed ({0}:{1})"
                          .format(datasources, info))
 
+class UnexpectedContentType(DependencyError):
+    def __init__(self, text, content_type, arg=None):
+        super().__init__("Expected content of type {}, "
```
AikoChou, line 80:
```suggestion
        super().__init__("Expected content of type {0}, "
```

revscoring/errors.py
```diff
@@ -75,6 +75,12 @@ def __init__(self, datasources, info=None, arg=None):
...
+                         "but the following can't be parsed "
+                         "(max 50 chars showed): {}"
```
AikoChou, line 82:
```suggestion
                         "(max 50 chars showed): {1}"
```

revscoring/errors.py
```diff
@@ -75,6 +75,12 @@ def __init__(self, datasources, info=None, arg=None):
         super().__init__("Query failed ({0}:{1})"
                          .format(datasources, info))
 
+class UnexpectedContentType(DependencyError):
+    def __init__(self, text, content_type, arg=None):
```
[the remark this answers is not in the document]
AikoChou: to fix the issue when testing exceptions picklability:
````
        uct = UnexpectedContentType("Test", "JSON")
>       pickle.loads(pickle.dumps(uct))
E       TypeError: __init__() missing 1 required positional argument: 'content_type'

tests/test_errors.py:75: TypeError
````

File: revscoring/errors.py (hunk 4 above)
Comment (AikoChou) [reply revscoring/errors.py:80]: use numbered indexes to be consistent with other error types

**Review by AikoChou — COMMENTED** [review]
I found a minor issue when I ran pytest for this

File: revscoring/errors.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (elukey) [reply revscoring/errors.py:79]: I'd like to keep the content_type as required field, if we default to None then there could be the possibility to have exception msgs like "Expected content of type None, etc..". Is it a limitation of pickle? Or maybe something else?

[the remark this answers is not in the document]
Comment (elukey) [reply revscoring/errors.py:79]: I think that the testing code should be:

uct = UnexpectedContentType(Dependent("Test"), "JSON")

I just updated the code change, lemme know if it fixes!

File: revscoring/errors.py (hunk 5 above)
Comment (elukey) [reply revscoring/errors.py]: Definitely I'll add the change to my patch!

File: revscoring/errors.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (AikoChou) [reply revscoring/errors.py:79]: It seems to be a problem of pickle.. if I remove the line `pickle.loads(pickle.dumps(uct))` and run the test, it will pass.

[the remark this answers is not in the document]
Comment (elukey) [reply revscoring/errors.py:79]: Ah ok makes sense, there may be some error in that Dependent("Test") class, we should figure out what's best to put in the test use case to make everything working.

**Comment by AikoChou** [comment]
@elukey I found a solution to keep the `content_type` as a required argument and remain pickleable. According to https://stackoverflow.com/questions/16244923/how-to-make-a-custom-exception-class-with-multiple-init-args-pickleable, we could override the `__reduce__` method in the exception:
```
def __reduce__(self):
    return (UnexpectedContentType, (self.text, self.content_type))
```
I tested it and it works!

PR #518, opened 2022-03-30
Improve error message when decoding a non-item revision from Wikidata

AikoChou:
Revscoring returns a very verbose and cryptic JSONDecodeError stacktrace
when it tries to score a non-item Wikidata revision with the item
quality model (for example, a User/Talk page change, etc..).
This change adds a new exception meant to capture this kind of error,
to wrap it in a clearer error message.

Co-authored-by: halfak <aaron.halfaker@gmail.com>,
Luca Toscano <ltoscano@wikimedia.org>
Bug: T302851

### DIFF
--- a/revscoring/about.py
+++ b/revscoring/about.py
@@ -1,5 +1,5 @@
 __name__ = "revscoring"
-__version__ = "2.11.1"
+__version__ = "2.11.2"
 __author__ = "Aaron Halfaker"
 __author_email__ = "ahalfaker@wikimedia.org"
 __description__ = ("A set of utilities for generating quality scores for " +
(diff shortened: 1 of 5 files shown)


halfak (approved):
This looks great.  I appreciate that you kept the pickle test.  That's critical when an error is thrown from within celery.  Everything that celery communicates must be pickleable.  This ensures that when an UnexpectedContentType error is thrown, we'll get it rather than a pickling error.  Nice work!

elukey:
Great job Aiko!

I just noticed that we should update the CHANGELOG file to reflect the fact that we'll release 2.11.2, could you send another code change when you have a moment?

@halfak hi! What is the process to release revscoring to pypi? Is it documented anywhere?

halfak:
Here's the script that used to perform the release after a version change was merged to master. https://github.com/wikimedia/revscoring/blob/master/scripts/deploy.sh  

I see that is disabled now, but it could be manually executed.

halfak:
I'll reach out via email to get it to you.

PR #521, opened 2022-05-19
Release 2.11.4

elukey:
The 2.11.3 release was created from a Python 3.9 venv, that should
work just fine but to be extra careful we created a new version
from a 3.7 venv.


kevinbazira (approved):
LGTM!

PR #531, opened 2022-12-22
ci: use bullseye image in CI

isaranto:
Using bullseye image for CI. In this PR we install the additional dependencies required to run tests etc using Debian Bullseye which is used by inference-services.
Changes:

- CI commands exist in Makefile so that we can easily transfer them to any platform
- updated dictionaries that have a hunspell package in debian. hunspell is a "superset" of myspell. Will also change these in inference-services image.
- fixed failing tests due to changes in language dictionaries
- added np.random.seed for test that failed randomly due to the values hardcoded in the test
- skipped test that is failing in CI for serbian language - to revisit with the python upgrade

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,32 @@
+name: "CI"
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  ci-merge:
+    runs-on: ubuntu-latest
+    container:
+      image: docker-registry.wikimedia.org/bullseye:20221204
+      options: --user root
+
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+
+      - name: Setup Wikimedia Buster image
+        run: |
+          apt-get update -y
+          apt-get install python3-pip python3-dev python3-setuptools build-essential python3-enchant g++ gfortran git \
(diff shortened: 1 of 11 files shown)


kevinbazira (commented):
LGTM

.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
+name: "CI"
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  ci-merge:
+    runs-on: ubuntu-latest
```
elukey, line 8: Nit: Is it a leftover? Should it be bullseye?

elukey:
Left a nit but the rest LGTM, merge when you prefer. Thanks for this work!

.github/workflows/ci.yml
isaranto: No it is not a leftover. The available options for github runners are macos, winows and ubuntu ([available runners](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on)).
So we use an ubuntu machine but run our things inside a container that runs the bullseye image.

elukey (dismissed):

elukey (approved):

PR #547, opened 2023-09-15
Release version 2.11.11

elukey:
The new version is required to be compatible with KServe 0.11

Bug: T346446


elukey:
The change has been tested on a Debian Bullseye container, running pytest.

isaranto (approved):
Works great!

